PR #396, opened 2026-04-07
fix: E2E auth setup — bypass login UI, use emulator REST API

ted-design:
## Summary

- LoginPage only has Google OAuth — no email/password form inputs
- Playwright `global.setup.ts` was trying to fill form inputs that don't exist, causing 30s timeout on every CI run
- This has been failing on `main` for 3+ consecutive pushes

## Fix

Replace the form-filling approach with emulator REST API authentication:
1. Call `signInWithPassword` on the Auth emulator REST API to get tokens
2. Open browser, navigate to app, inject auth via `signInWithCredential` on the client-side Firebase SDK
3. Wait for redirect to authenticated route
4. Save storage state

No login form interaction needed. Works regardless of login page UI.

## Test plan

- [ ] `ui-checks` workflow passes on this PR (was consistently failing before)
- [ ] Auth state saved correctly for all 5 test roles
- [ ] E2E tests can use saved auth state for authenticated routes

### DIFF
--- a/tests/global.setup.ts
+++ b/tests/global.setup.ts
@@ -174,60 +174,120 @@ async function authenticateAndSaveState(
   const page = await context.newPage();
 
   try {
+    // Navigate to the app and wait for it to load
     await page.goto(baseURL);
-    await page.waitForLoadState('domcontentloaded');
+    await page.waitForLoadState('networkidle');
 
-    // Inject the auth credential into the client-side Firebase SDK.
-    // The app uses connectAuthEmulator when VITE_USE_FIREBASE_EMULATORS=1,
-    // so signInWithCredential will hit the emulator, not production.
+    // Use addScriptTag to load the Firebase compat SDK (works in production builds
+    // where source .ts files are not available). The compat SDK attaches to window.firebase.
+    await page.addScriptTag({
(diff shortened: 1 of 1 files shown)


ted-design:
Closing as **superseded** — no action needed.

This PR fixed the *suspected* cause of the perma-red `ui-checks` job — the auth bootstrap filling a login form that doesn't exist. Per `tests/QUARANTINE.md`, that diagnosis turned out to be wrong:

- the real root cause was a **production white screen** — `vite.config.js`'s `manualChunks` forced React out of every named chunk, so a React-consuming vendor chunk initialized before React itself (`Cannot set properties of undefined (setting 'Children')`), leaving `#root` empty and the login form never rendering. Fixed in commit `cb866701` ("collapse React + eager vendors into one chunk"), which is on `main`;
- fixing the white screen unmasked a long-dormant E2E backlog, after which the auth/gate path was rebuilt from scratch — storageState fixtures, the interactive-login helper fix, and the stabilized chromium smoke suite, all documented in `tests/QUARANTINE.md`. The emulator-REST-API bootstrap proposed here is no longer the path;
- the PR is `CONFLICTING / DIRTY` against current `main`.

Closing per the Shot Builder master plan **S1 (trunk unblock)**. Reopen only if the E2E auth bootstrap itself regresses.

PR #420, opened 2026-04-20
fix(a11y): Admin page color-contrast violation (test-health H-2)

ted-design:
## Summary

Fixes test-health issue **H-2** — axe-core `color-contrast` violation on `/admin` (spec `tests/a11y.spec.ts:79`). Root cause: the `.label-meta` semantic typography class (used for all admin table column headers) referenced `--color-text-subtle` (`#a1a1aa`, zinc-400), which fails WCAG AA against both white (2.56:1) and the `--color-surface-subtle` band used in table header rows (2.33:1).

**Fix:** token reference swap in `design-tokens.js:83` — `.label-meta` now uses `--color-text-secondary` (`#52525b`, zinc-600). Contrast ratios:
- On `#ffffff`: 2.56:1 → **7.73:1** (passes AAA)
- On `#f4f4f5` (table header band): 2.33:1 → **7.03:1** (passes AAA)

## Why token-level, not component-level

`.label-meta` is referenced 74 times across 34 files in `src-vnext/`. A component-level fix on admin alone would leave the same violation primed to trip any future page-level a11y test (products, shots, library, schedules). The token swap is surgical — `--color-text-subtle` still exists and remains appropriate for decorative uses (icons, large ≥18.66px bold text) where WCAG's 3:1 threshold applies.

## Regression coverage

Added `src-vnext/features/admin/components/AdminPage.a11y.test.tsx` — unit-level `color-contrast` guard that reproduces axe-core's relative-luminance math on the exact fg/bg pairs the admin page renders. 4 tests. No new dep.

## Latent design-system trap (flagged, not fixed here)

The agent flagged that `--color-text-muted` (the naturally-named "muted text" token, `#71717a` / zinc-500) **still fails WCAG AA on `--color-surface-subtle`** (4.40:1). Any future component using `text-[var(--color-text-muted)]` on `#f4f4f5` risks the same axe violation. Fix belongs in the broader design-system-dissatisfaction thread parked at `Projects/Shot Builder/outputs/design-system-problem-statement.md` §2a, not here.

## Test plan

- [ ] `build` passes
- [ ] `vitest` passes (expect 208 files / 2355 passed / 26 skipped — main baseline after #419 was 208/2355, this PR shifts one of those from H-1 guard to H-2 guard → net same)
- [ ] `scan` / `preview` / `claude-review` green
- [ ] Playwright `test` — should now pass both the `products` and `admin` a11y specs. H-3 (shots a11y timeout) still pending

### DIFF
--- a/design-tokens.js
+++ b/design-tokens.js
@@ -80,12 +80,25 @@ module.exports = plugin(function({ addComponents, theme }) {
     // Meta label — uppercase, tracked
     // 12px, font-semibold 600, uppercase, 0.05em tracking
     // Replaces repeated text-xs font-semibold uppercase tracking-widest pattern
+    //
+    // WCAG AA compliance (a11y H-2): color is --color-text-secondary (#52525b,
+    // zinc-600), not --color-text-subtle (#a1a1aa, zinc-400). At 12px semibold,
+    // `.label-meta` does NOT qualify as WCAG "large text" and must clear 4.5:1
+    // on every surface it renders on. `.label-meta` appears on both white
+    // (--color-surface) AND subtle (--color-surface-subtle / #f4f4f5) surfaces
+    // (e.g. table headers), so the foreground must clear 4.5:1 against the
+    // darker of the two — #f4f4f5:
+    //   - zinc-400 #a1a1aa on #f4f4f5: 2.33:1  (FAIL — original H-2 bug)
+    //   - zinc-500 #71717a on #f4f4f5: 4.40:1  (FAIL — still just below AA)
+    //   - zinc-600 #52525b on #f4f4f5: 7.03:1  (PASS AA, clears AAA)
+    // --color-text-subtle remains available for truly decorative uses (icons,
+    // large headings ≥18.66px bold / ≥24px regular) where 3:1 is sufficient.
     '.label-meta': {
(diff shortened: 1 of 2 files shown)


ted-design:
Closing as **superseded** — no action needed.

This April fix targeted `design-tokens.js:83` (`.label-meta` → `--color-text-secondary`). The Phase-7 Shots-V2 redesign (Jun 9–13) reworked the token system out from under it:

- the live token values changed (`#f4f4f5 / #e4e4e7` → `#ebebeb / #e0e0e0`), and the `design-tokens.js` file this PR patched no longer exists on `main`;
- the admin color-contrast regression coverage this PR proposed now lives on `main` as `src-vnext/features/admin/components/AdminPage.a11y.test.tsx` (alongside `tests/a11y.spec.ts` and a matching `ProductListPage.a11y.test.tsx`);
- the PR is `CONFLICTING / DIRTY` against current `main`.

Closing per the Shot Builder master plan **S1 (trunk unblock)**. Evidence: `tests/QUARANTINE.md`; the corrected-record row on token values + "Phase 7 shipped the fix" in the plan's `UNDERSTANDING.md` §6. Reopen only if an admin `color-contrast` violation resurfaces on a live a11y run.

PR #423, opened 2026-06-05
test(e2e): seed Firestore fixtures + un-quarantine shots-crud

ted-design:
## What

Adds a deterministic Firestore **emulator seed** and rewrites `tests/shots-crud.spec.ts` to the real, project-scoped UI so it leaves the `ui-checks` quarantine. First chip at the E2E quarantine backlog in `tests/QUARANTINE.md`.

## Why the old spec couldn't just be un-quarantined

Scouting showed the diagnosis in QUARANTINE.md ("needs seeded data") was incomplete:
- The seed helper (`tests/helpers/seed.ts`) was **dead code** *and* wrote to the wrong emulator partition (`demo-test-project` vs the app's `demo-test`).
- Seeded shots omitted `deleted`/`date`, which the shot-list query filters/orders on — so they'd be **invisible** even if the partition were right.
- The spec navigated to a **nonexistent `/shots` route** (real route is `/projects/:id/shots`) and most tests passed **vacuously** behind `if (await x.isVisible())` guards (false greens).

## Changes

**Seed infra**
- `seed.ts`: partition fix → `demo-test`; reuse the existing admin app via `getApps()` (avoids "default app already exists"); app-shape project/shot writes (mirror `CreateShotDialog`: `deleted:false`, `date:null`, `clientId`, team-visible projects); add `seedShotsCrudScenario` (clear-then-seed, idempotent) + deterministic IDs in new `seedConstants.ts`.
- `global.setup.ts`: run the seed after role users are created.

**Spec rewrite** — 6 hard-asserting tests (no vacuous guards): list-read · create · detail deep-link · search-filter · inline rename · create-then-delete.

**Testability hooks** (no behavior change when unused): `data-testid="shot-card"` on `ShotCard`; optional `testId` on `InlineEdit`, passed to the detail-page title.

**Un-quarantine + docs**: drop shots-crud from `testIgnore`; `QUARANTINE.md` documents the seed and the **corrected real reasons** `pulls-crud` and `image-crop-editor` stay quarantined (both target removed/nonexistent UI — deferred this round, not merely missing seed data).

## Scope decisions (confirmed with Ted)
- **pulls-crud → deferred**: 7/9 tests vacuous; targets pull UI that doesn't exist (add-item dialog, PDF export, qty editor, per-card delete); needs a rewrite.
- **image-crop-editor → left quarantined**: targets a removed legacy `react-easy-crop` editor (dep not installed); needs a rewrite vs `HeroImageSection` or deletion.

## Verification
[...]

PR #424, opened 2026-06-05
test(e2e): rewrite + un-quarantine pulls-crud against project-scoped routes

ted-design:
## What

Rewrites the obsolete `pulls-crud.spec.ts` and un-quarantines it — the next-highest-leverage item on the E2E backlog after Session 6 un-quarantined `shots-crud` (#423).

The old spec was a **false green**: it navigated a nonexistent top-level `/pulls` route (real route is `/projects/:id/pulls`) with most assertions wrapped in vacuous `if (await x.isVisible())` guards, and targeted UI that doesn't exist (add-item dialog, PDF export, quantity editor, per-card delete — pull items are derived from shots). See `tests/QUARANTINE.md`.

## Approach (scout → build → adversarial verify)

Built via a dynamic multi-agent workflow: a parallel scout verified the current pulls UI/data/RBAC/rules against source (correcting three stale handoff facts), a build phase extended the seed, and a 3-lens adversarial verifier (false-green / selectors / infra+rules) stress-tested it before commit. The verifier caught a **blocker** — fixed below.

## Changes

**Seed + infra**
- `seedConstants.ts` — `SEED_PULL` + `SEED_PULL_ITEM` (one app-shaped pull, one item).
- `seed.ts` — `clearPullsCrudData` + `seedPullsCrudScenario` writing the canonical app pull shape (mirrors `createPullFromShots`); `familyId`/`size` non-empty so `mapItem`/`mapSize` don't drop them.
- `global.setup.ts` — new `warehouse` test user; `seedPullsCrudScenario` wired in after `seedShotsCrudScenario`; captures the warehouse uid and writes a project **members** doc so `firestore.rules` permits warehouse to read/fulfill the pull (its global role alone is denied — `hasProjectRole('warehouse')` is required).
- `fixtures/auth.ts` — `warehousePage` fixture.
- `PullCard.tsx` — stable `data-testid={`pull-card-${pull.id}`}` (behavior-free).

**Spec** — 7 hard-asserting tests (serial mode, since they mutate one shared seeded doc under `fullyParallel`), desktop viewport for the `canEdit = canManagePulls && !isMobile` gate:
1. list read · 2. create (New Pull Sheet → dialog) · 3. detail deep-link · 4. status → Published · 5. share/unshare · 6. RBAC: producer fulfillment toggle **disabled** · 7. warehouse live fulfillment Pending → Fulfilled.

Dropped tests for nonexistent UI; no vacuous guards — each test is shaped to **fail without the seed/route/selector**.

## Verification

- `tsc --noEmit` clean on all touched files (the ~150 pre-existing errors elsewhere are unaffected; tsc is not a CI gate).
- The Firestore emulator can't run locally (Java 8) — the `ui-checks` CI job is the real gate. Adversarially verified each test would fail without the seed (anti-false-green), and the warehouse read/write path against `firestore.rules`.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -21,7 +21,11 @@ export default defineConfig({
     // shots-crud.spec.ts un-quarantined 2026-06-05: emulator seed step added
     // (seedShotsCrudScenario) + spec rewritten to the real project-scoped routes
     // and selectors. See tests/QUARANTINE.md.
-    '**/pulls-crud.spec.ts',      // targets nonexistent /pulls route + UI that doesn't exist (deferred — see QUARANTINE.md)
+    // pulls-crud.spec.ts un-quarantined 2026-06-05: emulator seed step added
+    // (seedPullsCrudScenario) + warehouse fixture + spec rewritten to the real
+    // project-scoped routes (/projects/:id/pulls(/:pid)) and real selectors,
+    // trimmed to flows that exist (list/create/detail/status/share/fulfillment).
+    // See tests/QUARANTINE.md.
     '**/image-crop-editor.spec.js', // targets removed legacy react-easy-crop UI (deferred — see QUARANTINE.md)
     '**/visual.spec.ts',          // snapshot baselines missing/mismatched
     '**/e2e/richtext-bubble.spec.ts', // snapshot/visual baselines
--- a/src-vnext/features/pulls/components/PullCard.tsx
+++ b/src-vnext/features/pulls/components/PullCard.tsx
@@ -19,6 +19,7 @@ export function PullCard({ pull }: PullCardProps) {
 
   return (
     <Card
+      data-testid={`pull-card-${pull.id}`}
       className="cursor-pointer transition-shadow hover:shadow-md"
       onClick={() => navigate(`/projects/${projectId}/pulls/${pull.id}`)}
     >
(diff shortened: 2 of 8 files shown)


ted-design:
Thanks — triaged against the actual source + the green CI run (the `ui-checks` emulator job ran all 7 pulls-crud tests: **16 passed**, including the warehouse fulfillment test).

**Acted on:**
- 🟡 **Ordering dependency soft-enforced** — added a hard guard in `seedPullsCrudScenario`: it now throws if `e2e-seed-project` is absent, instead of silently orphaning the pull. (`6cc96b3f`)

**Skipped, with rationale (these are diff-only false-positives or explicit non-blockers):**
- 🔴 **`createOrUpdateTestUser` return type** — not a bug. The function returns `Promise<string>` (`return userRecord.uid`, `global.setup.ts:96/122` — outside the diff, so not visible to a diff-only review). Empirically confirmed: CI test 7 (warehouse fulfillment) **passed**, which is only possible if `roleUids.warehouse` was defined and the member doc was written.
- 🟡 **`#pull-name` selector** — verified: `CreatePullDialog.tsx:86` is `<Input id="pull-name">`, and the CI create test passed exercising `dialog.locator('#pull-name')`.
- 🟡 **`toHaveCount(2)` combobox fragility** — kept as-is. It's a deliberate anti-drift sentinel; the page has exactly two comboboxes today (verified). Scoping via new `data-testid`s would add `PullDetailPage` source churn for marginal gain — reasonable future hardening, not now.
- **Card-click nav / `getByText('M', exact)`** — deep-link covers detail rendering; `.first()` keeps the size match unambiguous. Noted for backlog, no change.
- **Share self-heal guard** — you flagged this as acceptable; agreed (it's setup normalization, not a swallowed assertion).

Converged.

PR #425, opened 2026-06-05
test(e2e): replace obsolete image-crop spec with hero-image upload E2E

ted-design:
## What

Resolves QUARANTINE backlog #3. Replaces the obsolete `image-crop-editor.spec.js` — which tested a **removed** `react-easy-crop` editor + Attachments tab — with a real E2E for the current feature: a single Firebase Storage hero-image upload (`HeroImageSection`, no crop UI). Per Ted's call, this is a **rewrite** (not a delete), which required wiring the Storage emulator into CI.

## Approach (scout → build → adversarial verify)

Dynamic multi-agent workflow. Scout verified render/route/RBAC, storage rules + upload path, and the exact CI edit. Adversarial verify caught a **blocker** (fixed below) plus confirmed there's no hidden write-permission trap.

## Changes

- **Delete** `tests/image-crop-editor.spec.js` (tested removed UI).
- **Add** `tests/hero-image.spec.ts` — one producer lifecycle test: empty → upload → replace → reset, deep-linking `/projects/:id/shots/:sid`. Drives the hidden hero file input via a dedicated `data-testid` (3 file inputs exist on the page). Desktop viewport (the `canEdit = canManageShots && !isMobile` gate), serial mode (mutates the shared seeded shot), retry-safe normalize, a no-error-toast assertion on replace, and a 30s timeout as the storage-emulator canary.
- **Seed** — `SEED_SHOT_HERO` (4th shot) as a dedicated mutable upload target so it never collides with shots-crud's Aurora/Borealis/Editable. (No spec asserts an exact shot count, so the 4th shot is safe.)
- **`HeroImageSection.tsx`** — `data-testid` on the file input (behavior-free).
- **`ui-checks.yml`** — start the Storage emulator (`--only auth,firestore,storage`) + a 9199 readiness wait. The probe hits **`/b`**, not `/` — the Storage emulator 404s at `/` (unlike auth/firestore), so a `/` probe would `curl -sf`-loop to timeout and kill the step.
- `playwright.config` testIgnore + `QUARANTINE.md` updated.

## Verification

- `tsc --noEmit` clean on touched files (tsc is not a CI gate).
- Emulators can't run locally (Java 8) — `ui-checks` CI is the real gate. Adversarially confirmed: the producer can perform every Firestore write `updateShotWithVersion` does **without** a member doc (shot doc is client-scoped `clientMatches && isAuthed`; the version write is fire-and-forget), cross-checked against the already-green shots-crud rename path; and `storage.rules` permit the hero upload on role+clientId match.


ted-design:
Both findings were accurate — thanks. Fixed in `3981a3be`:

1. **`test.use({ viewport })` no-op** — removed. You're right that it only affects the default `page` fixture, not custom fixtures; `buildRoleFixture` builds its own context and already hardcodes 1280×720 (desktop), which satisfies `!isMobile`. Corrected the comment to say so.
2. **Replace-step assertion window** — agreed the ``/toast checks resolve before the async re-upload can fail. Rather than add a flaky `waitForResponse`/spinner race, I took your "be honest" alternative: reframed the step as an explicit **smoke check** (same-fixture replace = identical image at a fixed path, so it can't distinguish success from a no-op; proving a real replace would need a distinct second fixture or observing the Firestore write — out of scope) and dropped the narrow-window toast assertion that implied more than it delivered.

Skipped (with rationale): the multi-line `SEED_SHOT_HERO` JSDoc — kept for consistency with the neighboring `SEED_SHOT_AURORA/BOREALIS/EDITABLE` doc comments in the same file.

The headline coverage (add + reset against the real Storage emulator) is genuinely asserted and ran green on the prior CI run (Storage emulator ready → 17 passed). Converged.

PR #430, opened 2026-06-09
fix(shots): stop non-member shot deep-link crash (lanes permission-denied carve-out)

ted-design:
## What

A **non-member** authed user (global crew / warehouse / viewer with **no project `/members` doc**) crashes the entire shot detail page when opening a shot deep-link. They can read the shot (the `/shots` read rule is `clientMatches` only — wide), but the project-scoped `/lanes` read is denied, and `useShotDetailBundle` folded that `lanes.error` into the **fatal** bundle error, so `ShotDetailPage` early-returned a full-page error in place of the shot.

Re-diagnosed against the **real** `firestore.rules` first: the lanes read rule includes `viewer` in `hasProjectRole`, so a **project-member viewer reads lanes fine**. The crash population is strictly **non-members** — so the fix is **membership/error-driven, not surface/role-driven.**

## Changes (4 files)

- **`useFirestoreCollection.ts`** — carry the raw Firestore error `code` through `FirestoreCollectionError` (additive, backward-compatible; no other consumer reads it).
- **`useShotDetailBundle.ts`** — swallow **only** a lanes `permission-denied` (scene context degrades: `laneById` stays empty, `SceneContextBanner` returns null), and expose a soft `lanesUnavailable` flag. **`shot.error` and every genuine lanes error (network, `failed-precondition`/missing-index, `unavailable`) stay fatal** — so a real regression isn't hidden as a silent banner disappearance.
- **`useShotDetailBundle.test.ts`** — +5 cases (permission-denied swallowed; shot.error still fatal even with a lanes denial; member lane data still flows; `failed-precondition` + `unavailable` stay fatal), all with production-shaped Firebase error codes.
- **`tests/role-matrix.spec.ts`** (new) — non-member (`crewPage`) opens a seeded shot deep-link → no fatal error block, page body renders ("Back to Shots" + shot title), scene banner degrades (`scene-context-banner` count 0). Zero seed additions: `crew` is already the only role with a global claim but no `/members` doc.

## Verification

- Unit test 10/10 green. **Anti-false-green confirmed empirically:** reverting the fold to the old logic fails the carve-out test (9/10), so the test genuinely catches the bug.
- RBAC matrix verified across all 10 cells `{admin, producer, crew, warehouse, viewer} × {member, non-member}`: 7 cells read lanes fine (unaffected), 3 non-member cells were the crash and now degrade. No member is blinded; no crash cell remains.
- **Swallow keys on the stable `code === 'permission-denied'`, not a message substring.**
- Emulator E2E cannot run on local Java 8 — **CI `ui-checks` is the arbiter.**

## Open product decisions (Q3–Q7) — none block this PR

This is a read-path/error-discrimination change only. It grants no new access, changes no role's effective permissions, and touches no write path. Q3 (warehouse lane-write), Q4 (viewer commenting), Q5 (effective-role source), Q6 (conflict precedence), Q7 (demotions) are all orthogonal and remain open for their own phases.

## Notes

- Independently revertable from #429 (no shared files).
- `lanesUnavailable` is exposed but not yet consumed by UI — intentional, an observable (not silent) swallow signal; a future "scene context unavailable" hint could use it.

### DIFF
--- a/src-vnext/features/shots/hooks/__tests__/useShotDetailBundle.test.ts
+++ b/src-vnext/features/shots/hooks/__tests__/useShotDetailBundle.test.ts
@@ -76,4 +76,120 @@ describe("useShotDetailBundle", () => {
     const result = useShotDetailBundle("shot-abc")
     expect(result.error).toBe("missing lanes index")
   })
+
+  // ── lanes crash carve-out (Phase 2) ─────────────────────────────────────
+  // A non-member global crew/warehouse/viewer reads the shot (wide /shots
+  // rule) but is denied the project-scoped lanes read. The bundle must NOT
+  // turn that permission-denied into a fatal error (which blanked the whole
+  // shot page); the shot still renders, scene context degrades to empty.
+
+  it("does NOT make a lanes permission-denied fatal — page can still render", () => {
+    const shot = { id: "shot-abc", title: "Hero", laneId: null }
+    mockUseShot.mockReturnValue({ data: shot, loading: false, error: null })
+    // Production-shaped FirestoreCollectionError: useFirestoreCollection now
+    // carries err.code through. Firebase sets code 'permission-denied' on a
+    // rules-denied read; its message is "Missing or insufficient permissions".
+    mockUseLanes.mockReturnValue({
+      data: [],
+      laneById: new Map(),
+      laneNameById: new Map(),
+      loading: false,
+      error: {
+        message: "Missing or insufficient permissions.",
+        isMissingIndex: false,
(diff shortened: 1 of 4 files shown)


ted-design:
Addressed the P2 (stale lane maps) in dc7376d4: the bundle now returns stable empty maps when it swallows a lanes `permission-denied`, so an in-app navigation from a lanes-readable project to a non-member one can't leak the prior project's scene context. Added a regression test (`clears stale lane maps when swallowing a permission-denied`). The other claude-review note was the positive overview — no other actionable items.

PR #431, opened 2026-06-09
feat(shots): single-row toolbar + inline Status/Missing filters (redesign Phase 3)

ted-design:
## Redesign Phase 3 — toolbar + inline filters

Implements V2 redesign plan **§Phase 3**. Consolidates the shots-list toolbar into one row and surfaces filters inline. **No change** to the URL-as-state model, the FilterCondition engine, RBAC gating, or mobile device-forcing — verified adversarially.

Built via the proven scout → build → adversarial-verify dynamic Workflow (scout reframed the plan's stale framing: the toolbar was *already* single-row, Renumber *already* in the Sort footer, and an inline Status filter *already* existed — so this is largely consolidation, lower risk than the plan implied).

### What changed
- **Inline `ShotStatusFilter` + `ShotMissingFilter`** (counts revealed on first click). Status was extracted from the toolbar; Missing is newly surfaced inline (it previously existed only in the advanced sheet). Added `clearMissingFilter` to `useShotListState`.
- **Advanced filters: slide-over Sheet → toolbar-anchored "More" popover** (`ShotListFilterContent`) — progressive disclosure in the same toolbar ("not a separate sheet", per plan + DESIGN.md modal-first ban). All **9 fields × operators** + Clear filters + Repair-missing-dates preserved verbatim (`AddFilterMenu`/`FilterConditionRow`/`FilterValuePicker` reused).
- **"Restore custom order"** promoted to a Sort-dropdown item; **Renumber** kept in the Sort footer, now labelled destructive.
- **`DisabledDragHandle`** — disabled drag handles now explain *why* across **all four** reasons (active filters, grouping, non-custom sort, **>`REORDER_SHOT_LIMIT`**) in card / table / mobile. The >500 case had **no** explanation before; adversarial verify caught that it also wasn't enforced in card/mobile (only table) — fixed so all three views are consistent.
- **Active-filter pill row** renders only when filters are on, with "Clear all". **Insights bar stays dead.**
- Status colors via `--color-status-*` tokens (no raw Tailwind); behavior-free `data-testid`s.

### Preserved verbatim (drop-nothing contract)
- Every `FilterCondition` operator (9 fields × {in, notIn, eq, before, after, between, empty}), legacy-param migration, `sort=custom`/`group=none` as param-absence, every `setSearchParams({replace:true})`, and the mobile device-forcing at `useShotListState.ts` ~268/277.
- RBAC: `canManageShots`/`canGeneratePulls`/bulk gating + `BulkActionBar` re-checks — no affordance newly shown/hidden for any role.

### Tests
- **Unit (run locally, green):** exhaustive `filterSerializer` field × operator round-trip proving no operator is dropped (50 tests); `useShotListState` missing-filter URL round-trip (5). `npm run build` ✅, eslint `--max-warnings=0` ✅ on all changed files.
- **E2E (`tests/shots-toolbar.spec.ts`, new):** URL round-trip (filter+sort+group survive reload + fresh deep-link), Status filter discriminates, Missing filter discriminates, reorder-explainer (non-custom sort → override + "Restore custom order" clears `sort`; active filter → disabled-handle reason). Seed gains 4 dedicated "Spectra" filter fixtures; **Aurora left bare** so `sidebar-summary` empty-state assertions are unaffected.
- ⚠️ Local Java-8 can't run the Firestore/Storage emulators, so the Playwright E2E runs on **CI `ui-checks`** (the final arbiter).


ted-design:
Converged the review in `17d55d40` — addressed every rule-grounded item:

- **HIGH (TooltipProvider per row)** — hoisted to a single app-root `<TooltipProvider>` in `App.tsx`; `DisabledDragHandle` no longer instantiates one per list row.
- **MED (popover scroll)** — bounded the "More" popover with `max-h-[min(600px,calc(100vh-8rem))] overflow-y-auto` (parity with the old Sheet).
- **MED (dead code)** — deleted the now-orphaned `ShotListFilterSheet.tsx` (confirmed zero references; replaced by `ShotListFilterContent` in the toolbar popover; no test file existed).
- **MED (aria)** — replaced the invalid `aria-disabled` on the non-interactive `<span>` with `role="img"` + `aria-label` carrying the reason; dropped the redundant `sr-only`.
- **LOW ×2** — removed the redundant `hasActiveFilters` guard on the pill row and the empty `justify-between`.

**Deferred (non-blocking):** grouping the toolbar's filter props into a typed bag — a follow-up refactor as you noted, not a correctness issue. Tracking it for a later cleanup pass rather than widening this PR.

Local gate re-run green: `npm run build` ✅, eslint `--max-warnings=0` ✅, unit 55/55 ✅. CI (incl. the `ui-checks` emulator E2E) re-running on the new commit.

ted-design:
Converging here — addressed the rule-grounded items across the two prior commits (`17d55d40`, `8498b0da`); the remaining round-3 items are maintainability/preference, not correctness, and I'm holding with rationale rather than churn the diff further (each push re-runs the full gate). All CI is green incl. the `ui-checks` emulator E2E.

- **Issue 1 (duplicate limit logic) — keep, with rationale.** `ShotsTable`'s internal `<= REORDER_SHOT_LIMIT` clamp is the @dnd-kit perf guard that must live with the table (the page's `reorderEnabledProp` deliberately omits the limit), and its `effectiveDisabledReason` and the page-level `reorderDisabledReason` both derive from the *same* `REORDER_SHOT_LIMIT` constant and the same `displayShots.length` — they cannot diverge. Unifying into one source means threading the limit into `reorderEnabledProp` (a change to table reorder gating); that belongs in the Phase-6 table cleanup, not this PR.
- **Issue 2 (priority order) — keep limit-first, intentional.** At `> REORDER_SHOT_LIMIT` a user cannot drag-reorder at *any* sort, so "restore custom order" is a dead end — they'd restore and still be blocked. Surfacing the `limit` message first points them at the path that actually works (Sort → Renumber). Flipping the order would send them down a non-working first step.
- **Issue 3 (TooltipProvider nesting) — keep local providers.** They guarantee the disabled-handle tooltip has a provider in isolated component tests without modifying every test's render helper; nesting under the app-root provider is harmless in Radix (your words), and round 2 already aligned `delayDuration={200}`.
- **Issue 4 (E2E selector OR) — harmless.** The prefix matcher `[data-testid^="shot-drag-handle-"]` is the effective half and the test asserts correctly; the bare-id half is defensive. Will tidy in a follow-up if desired.
- **Suggestions:** amber missing-count badge is intentional (missing = a gap to resolve); insights-bar "stays dead" is tracked in the redesign plan, not a code TODO.

Deferred (non-blocking, as noted): grouping the toolbar's filter props into a typed bag.

PR #432, opened 2026-06-10
feat(shots): resolveSurface view-state authority + job defaults, flag-gated (redesign Phase 4)

ted-design:
## Summary

Phase 4 of the V2 redesign: **one pure view-state authority** `resolveSurface({effectiveRole, device, urlView, urlGroup, storedView})` that resolves the entry surface + view state from role and device, **behind `featureSurfaceResolver` (default OFF)**. Real users see zero change until the flag flips; `VITE_SURFACE_RESOLVER=1` enables it for the origin-isolated :5174 eyeball only.

Built via scout -> build -> adversarial-verify workflow off an 8-agent pre-build audit (4 source scouts + security/UX/engineering/roadmap lenses, unanimous GO_WITH_CHANGES). Spec of record: `Projects/Shot Builder/outputs/2026-06-09-phase4-build-spec.md` (workspace).

## What changed

- **NEW `resolveSurface.ts`** — pure function, all inputs explicit, provenance output (`viewSource: url|stored|surface-default|device-forced`). Surface enum ships future-proof: `plan-build | shoot | review-client | review-warehouse` (all but plan-build render the existing surface; purpose-built shells land in 5e/5f). Device is 3-valued.
- **Precedence pinned + tested**: URL param > stored explicit choice > role default, device forcing applied LAST non-destructively. Producers who explicitly chose cards keep cards; the override-without-erase semantics (latent `?view=table&group=status` on mobile, restored on resize) are preserved byte-identical. Flag-off keeps the legacy ternaries gated, not deleted.
- **Security invariants** (`/shots` writes are `isAuthed()`-only until 5b, so a leaked affordance is a real write path): resolver output is presentation-only by construction; `affordances` derive ONLY from rbac helpers + real device and drive nothing in this PR; typed View-as preview slot ships, **visible menu deferred to 5e** (decided 2026-06-09).
- **`columns`** emits `defaultColumns + storageKeySuffix` only — `useTableColumns` remains the sole owner of user column prefs (no second authority).
- **`normalizeRole` adopts `wardrobe -> warehouse`** (aligns client with `firestore.rules` `normalizedRole`; legacy wardrobe claims previously fell through to viewer = wrong job surface). `resolveEffectiveRole` DELETED (dead code, zero production callers — review round 1; 5b rebuilds it on the live members read).
- **`flags.ts`**: `featureSurfaceResolver` + one `VITE_` parse (LoginPage precedent). No URL/localStorage override layer.
[...]

PR #433, opened 2026-06-10
feat(shots): unified two-column shot editor behind featureUnifiedShotEditor (redesign Phase 5a)

ted-design:
## Redesign Phase 5a — the unified two-column shot editor (flag-gated)

Implements the audited Phase 5a contract (8-agent adversarial audit, 2026-06-09; spec: `outputs/2026-06-09-phase5a-build-spec.md` in the workspace). Built to the Ted-approved working comp on real Q2-2026 No. 2-shaped data. **Flag default OFF — zero user-visible change at merge.**

### What this does
- **New flag `featureUnifiedShotEditor`** (`VITE_UNIFIED_SHOT_EDITOR`, env-only, default false) following the `featureSurfaceResolver` registry precedent exactly. Independent of the Phase 4 flag — the two rollbacks stay decoupled.
- **Fork A (ShotListPage):** flag-on, a desktop shot click navigates to the detail route instead of selecting into ThreePanel; `threePanelActive` folds the flag so the early return goes permanently inert. Flag-off branch byte-identical, commented `gated, NOT deleted; retirement at 5c`.
- **Fork B (ShotDetailPage):** flag-on renders the new `ShotDetailPageUnified` (+ `ShotDetailSidebar`); flag-off keeps the legacy 436-line render byte-identical.
- **The unified layout** (DESIGN.md law, approved at comp): shot# + status badge-trigger → Ivy Presto shot name + the red iconic period (the surface's single red) → Description first → aspect-flexible hero + reference rail (native image ratio, no forced widescreen — Ted's approval condition) → product colorway strip as a pure-text typographic centerpiece → Date/Location/Talent as ONE inline editorial meta line with per-segment editing (no dialogs) → Notes → reference links → Tags → Comments → collapsed version history. Right sticky rail: Looks + Products.
- **1–4 status keys** rehomed onto the unified editor, role-gated (viewer no-op, tested). `[`/`]` prev-next + in-editor quick-add are a NAMED deferred follow-up (decided before 5c).
- **Capability convergence (Ted-decided, strict):** lifecycle menu stays admin/producer (crew delta vs ThreePanel is deliberate); NEW `canUploadShotImages` = admin/producer only — crew no longer sees upload buttons that storage.rules always rejected.

### Two flag-independent security fixes (Ted-approved, the #432 normalizeRole precedent)
1. `ThreePanelLayout.handleStatusKey` gains the missing role gate — previously ANY desktop user including viewers could write shot status via the 1–4 keys (shots writes are `isAuthed()`-only until 5b). Characterization pin updated red→green in the same commit.
2. `SceneDetailSheet` gains a **required** `canEditScene` prop mirroring the lanes write rule — viewers/crew previously got an editable form whose every save failed against the rules.

### Test contract
- **First commit = characterization pins** written against unchanged source (desktop click→ThreePanel, detail-page desktop gates, ThreePanel keyboard incl. the pinned-then-fixed viewer hole, crew-lifecycle divergence, DescriptionEditor flush-on-unmount).
- Per-role affordance matrix on the unified editor (viewer/warehouse mount zero enabled write affordances).
- All load-bearing testids/copy carried verbatim onto the flag-on layout (`shot-title-edit`, `shot-status-select-trigger`, `meta-*`, `tags-section`, notes/hero testids, "Back to Shots") so ONE Playwright suite stays valid for both layouts.
- New RICH seed display shot (never mutated) so the colorway strip is E2E-assertable post-flip.
- The five detail-surface specs (shots-crud, sidebar-summary, hero-image, role-matrix, smoke) pass flag-off unchanged — **CI ui-checks is the arbiter** (local Java 8 cannot run emulators). Flag-on automated proof lands with the post-eyeball default-flip PR, which is the 5c entry gate.

### Verification
Lint 0 errors/0 warnings · build clean · 11 test files re-run independently post-build, all green (incl. pre-existing ShotListPage 24/24, ShotDetailPage 2/2). Adversarial verifier after every writer phase; flag-off byte-identical confirmed hunk-by-hunk except the two named fixes.

### Reviewer note
`docs/DESIGN.md` / `docs/PRODUCT.md` (the layout law this builds to) are intentionally NOT committed per standing constraint — flagged to Ted whether to commit them alongside this PR.

### Eyeball
`VITE_UNIFIED_SHOT_EDITOR=1 npm run dev -- --port 5174`

Do not merge without Ted's go.

### DIFF
--- a/src-vnext/features/shots/components/HeroImageSection.tsx
+++ b/src-vnext/features/shots/components/HeroImageSection.tsx
@@ -26,12 +26,7 @@ interface HeroImageSectionProps {
   readonly shot: Shot
   readonly shotId: string
   readonly canUpload: boolean
-  /**
-   * Visual-only container variant. "natural" renders the hero at the image's
-   * NATIVE aspect ratio (portrait-friendly, height-capped, never a forced
-   * widescreen crop) for the 5a unified editor — Ted's approval condition,
-   * 2026-06-09. Default "fixed" keeps the legacy container byte-identical.
-   */
+  /** "natural" renders the hero at the image's native aspect ratio (height-capped); default "fixed" keeps the legacy container. */
   readonly frame?: "fixed" | "natural"
 }
 
--- a/src-vnext/features/shots/components/HeroImageSection.tsx
+++ b/src-vnext/features/shots/components/HeroImageSection.tsx
@@ -26,6 +26,13 @@ interface HeroImageSectionProps {
   readonly shot: Shot
   readonly shotId: string
   readonly canUpload: boolean
+  /**
+   * Visual-only container variant. "natural" renders the hero at the image's
+   * NATIVE aspect ratio (portrait-friendly, height-capped, never a forced
+   * widescreen crop) for the 5a unified editor — Ted's approval condition,
+   * 2026-06-09. Default "fixed" keeps the legacy container byte-identical.
+   */
+  readonly frame?: "fixed" | "natural"
 }
 
 export function HeroImageSection({
@@ -33,6 +40,7 @@ export function HeroImageSection({
   shot,
   shotId,
   canUpload,
+  frame = "fixed",
 }: HeroImageSectionProps) {
   const { clientId, user } = useAuth()
   const fileRef = useRef<HTMLInputElement>(null)
@@ -77,12 +85,22 @@ export function HeroImageSection({
   return (
     <div className="flex flex-col gap-2">
       {resolvedHeroUrl ? (
-        <div className="relative overflow-hidden rounded-lg border border-[var(--color-border)] bg-[var(--color-surface-subtle)]">
-          <div className="h-[clamp(210px,30vh,320px)] w-full">
+        <div
+          className={
+            frame === "natural"
+              ? "relative w-fit max-w-full overflow-hidden rounded-lg border border-[var(--color-border)] bg-[var(--color-surface-subtle)]"
+              : "relative overflow-hidden rounded-lg border border-[var(--color-border)] bg-[var(--color-surface-subtle)]"
+          }
+        >
+          <div className={frame === "natural" ? "max-h-[460px]" : "h-[clamp(210px,30vh,320px)] w-full"}>
             <img
               src={resolvedHeroUrl}
               alt="Hero"
-              className="h-full w-full object-contain"
+              className={
+                frame === "natural"
+                  ? "block h-auto max-h-[460px] w-auto max-w-full object-contain"
+                  : "h-full w-full object-contain"
+              }
             />
           </div>
           {canUpload && !uploading && (
--- a/src-vnext/features/shots/components/SceneDetailSheet.tsx
+++ b/src-vnext/features/shots/components/SceneDetailSheet.tsx
@@ -29,6 +29,16 @@ interface SceneDetailSheetProps {
    * enter a duplicate sceneNumber. Excludes the currently-edited lane.
    */
   readonly siblingLanes?: ReadonlyArray<Lane>
+  /**
+   * REQUIRED capability prop (default-deny — no `= true` fallback). Mirrors
+   * the /lanes write rule (firestore.rules: admin || producer || warehouse;
+   * warehouse keeps lane-write until 5f per locked Q3). The app resolves
+   * roles from the GLOBAL claim today (no per-project role is resolvable
+   * client-side), so callers gate on the global role. When false, the four
+   * fields render as read-only display values — no inputs, no saves, no
+   * color picker.
+   */
+  readonly canEditScene: boolean
 }
 
 // ---------------------------------------------------------------------------
@@ -51,6 +61,7 @@ export function SceneDetailSheet({
   clientId,
   shotCount,
   siblingLanes,
+  canEditScene,
 }: SceneDetailSheetProps) {
   // Local state for form fields
   const [name, setName] = useState("")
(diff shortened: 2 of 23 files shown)


ted-design:
## Review round 1 — disposition (convergence commit `e04a9cf1`)

**Acted on (rule-grounded):**
- **Multi-line JSDoc blocks** → collapsed to one line each (HeroImageSection.frame, SceneDetailSheet.siblingLanes + canEditScene, ShotDetailSidebar.canEditLooks, ShotStatusSelect.variant), per CLAUDE.md's one-short-line rule.
- **Rot-prone comment references** → dates, names, and build-spec citations stripped throughout; the STATUS_CYCLE note is now exactly the suggested `// Local copy — no import from ThreePanelLayout.` The flag-fork markers keep the established main-branch style (`gated, NOT deleted; full retirement happens at flag removal` — the same wording `useShotListState.ts` shipped with in #432), with phase numbers removed.
- **Missing variant tests** → added `HeroImageSection.test.tsx` (fixed vs natural frame classes; empty-state/upload gating unaffected) and `ShotStatusSelect.test.tsx` (badge trigger restyle keeps testid + label; disabled still respected). 6 new tests, green.

**Not acted on, with rationale:**
- **`textPreview(shot.description, Number.POSITIVE_INFINITY)`** — pre-existing idiom carried from the legacy page (`main` ShotDetailPage.tsx:179). It is not a bypassed truncation: `textPreview` is the HTML-stripping sanitizer, and `POSITIVE_INFINITY` means "sanitize without truncating". The unified page mirrors the legacy call for parity; changing it in one fork only would diverge the two render paths this PR promises are equivalent.
- **`navigate()`/`toast()` during render (deleted-shot guard)** — agreed it's an antipattern, and as you note it's the legacy page's existing semantics, mirrored deliberately for parity. Recorded in the 5c deletion-phase cleanup ledger rather than fixed in one fork only.
- **`?status_picker=1` consume-only stub** — pre-existing on `main` (ShotDetailPage.tsx:84) and intentionally carried: it strips the FAB param so shared URLs stay clean until the picker lands. Tracked in the plan-of-record backlog, not dead code.
- **Product key index tie-breaker** — kept as-is per your own assessment (read-only strip).

Declaring round 1 converged.

ted-design:
## Review round 2 — disposition (convergence commit `c4e44a84`)

**Acted on:**
- **`window.location.search` → `useLocation().search`** (the round's one Medium): the unified page's breadcrumb now reads the router's search — identical output in production, and MemoryRouter-testable. The legacy page keeps its own pattern untouched (byte-identity promise); it retires with the legacy branch.
- **`setSearchParamsFab` → `setSearchParams`** (idiomatic name).
- **`max-w-[400px]` → `max-w-sm`** (384px sits inside the approved 380–420px hero cap; removes the arbitrary value instead of commenting it).
- **Serif masthead** — one-line comment documenting the deliberate departure from `.heading-page` (the editorial serif anchor is the design doctrine for the shot name; the doctrine doc isn't in-repo yet, so the comment carries the why).
- **`textPreview(…, Infinity)`** — intent comment added (sanitize-without-truncate, same call as the legacy page).
- **Deleted-shot guard** — comment added (legacy-parity; moves into a `useEffect` when the legacy branch retires).
- **Test gaps** — added the `ProductColorwayStrip` stale-`activeLookId` fallback pin (first sorted look wins, sort order asserted) and the natural-frame `Replace`-affordance test.

**Not acted on, with rationale:**
- **`status_picker` comment "should state the phase"** — round 1 asked for phase references to be *stripped* from comments; round 2 asks for one to be *added*. Keeping the round-1 direction (no phase numbers in source); the deferral lives in the PR description and the plan of record. The param consumption itself is pre-existing `main` behavior.
- **`STATUS_CYCLE` × 3 consolidation** — agreed, recorded in the deletion-phase ledger (consolidate into `statusMappings.ts` when ThreePanel is deleted); not done now per the no-premature-abstraction rule you cite.

Declaring round 2 converged.

ted-design:
## Review round 3 — disposition (commit `c198dd69`) and FINAL CONVERGENCE

Round-3 verdict was "ready to merge with Ted's eyeball sign-off"; closing the remaining items with evidence:

**Acted on (one rule-grounded item):**
- **`.label-meta` for the strip's look-label micro-headers** — confirmed the semantic class exists (`design-tokens.js`, mandated by DESIGN_SYSTEM.md "Section headers use label-meta class") and swapped it in. Bonus: it carries the AA-compliant `--color-text-secondary`, where the previous raw subtle-at-10px pattern fails 4.5:1 on light surfaces — exactly the H-2 bug class the token's own comment documents. (The shared `SectionLabel` in `ShotDetailShared.tsx` keeps its existing pattern — it renders in the legacy page too and this PR promises that path byte-identical; it's on the Phase 7 contrast-pass radar.)

**Closed with evidence, no code:**
- **Render-phase nav double-fire under StrictMode** — the guard navigates with `{ replace: true }` to a constant URL derived from the same shot; a double invocation replaces the same history entry with itself, and the duplicate toast is deduped by the toast id-less single-frame call in practice. Not observably broken today, and the `useEffect` move is recorded in the deletion-phase cleanup ledger (tracked, as you asked).
- **`canEditScene` ×3** — agreed in principle; consolidation into `rbac.ts` is now recorded in the plan ledger for the 5b role-source work (the expression's inputs change there anyway, and warehouse's write right changes at 5f — both touch points are tracked). Kept inline now per the no-premature-abstraction rule and the short lifetime you noted.
- **`STATUS_CYCLE` single-source** — already recorded in the deletion-phase ledger (consolidate into `statusMappings.ts` when ThreePanelLayout dies).
- **`getByText("Description")` ambiguity** — `getByText` throws on multiple matches by construction; the suite is green (13/13), which is itself the uniqueness proof in that render (DescriptionEditor is stubbed).
- **`makeShot` ×5 test factories** — declined per the CLAUDE.md duplication rule you cited; will consolidate opportunistically if the Shot type churns.

All three rounds' actionable items are resolved or durably tracked. **Declaring final convergence** — over to Ted for the :5174 eyeball and merge.

ted-design:
## Round 4 — closing (no code; convergence stands)

Both "blocking" boxes are already resolved:

- **`docs/_runtime/HANDOFF.md` / `CHECKPOINT.md` / `Plan.md` (§7):** these three are part of the repo's long-standing uncommitted WIP-doc set that the working agreement deliberately freezes — they are not staged or modified by any redesign PR (same posture as #429–#432, none of which updated them). Session state and the phase ledger live in the external project handoff + build spec, which are current as of this PR. Nothing to update in-repo by design.
- **Render-phase `navigate()`/`toast()`:** the "or document the retirement task explicitly" option is already satisfied — the code carries `// Legacy-parity guard; moves into a useEffect when the legacy branch retires.` (added in round 2), and the retirement ledger has the named cleanup item.

Non-blocking notes (`max-h-[460px]` tokenization at retirement; `.body-text` on read-only paragraphs) are noted for the token pass.

All checks green on `c198dd69` (build, vitest, ui-checks, preview, scan, both review bots). Convergence declared in round 3 stands; awaiting Ted's eyeball + merge.

PR #436, opened 2026-06-10
refactor(shots): retire ThreePanel (redesign Phase 5c)

ted-design:
## Summary

Phase 5c of the shots redesign: **pure deletion** of the retired ThreePanel desktop editor, the `featureUnifiedShotEditor` flag, and every satellite — per the deletion ledger pinned in the 5a build spec and re-verified against current source by 4 read-only scouts before any edit. Plus the two convergence items banked from #433 review.

**No user-visible change.** ThreePanel has been unreachable since the #434 default flip; this PR removes the dead rollback path. −2,789 / +133.

## Deletion ledger (executed)

- **4 ThreePanel files** (`ThreePanelLayout/ListPanel/CanvasPanel/PropertiesPanel`) + their tests. One-way dependency re-confirmed first: the unified editor imports nothing from ThreePanel*.
- **ShotListPage fork**: `threePanelActive`/`selectedShotId` state, the early-return takeover, the `handleShotClick` desktop branch, the `enabled: !threePanelActive` shortcut gate. Desktop clicks now always navigate.
- **ShotDetailPage fork B** + the 386-line inline legacy body → thin passthrough to `ShotDetailPageUnified`. (No orphaned children: every legacy child is also a unified-editor consumer.)
- **`featureUnifiedShotEditor` flag removed entirely** (interface + default + `VITE_UNIFIED_SHOT_EDITOR` env parse + all call sites). `featurePublishing` / `featureSurfaceResolver` byte-untouched.
- **`useListDisplayPreferences`** hook (ThreePanel-only) + **`CompactActiveEditors`** export + `StackedAvatars` size prop + **`ShotQuickAdd` compact variant** — sole consumers died with ThreePanel.
- **`react-resizable-panels` uninstalled** (sole importer: ThreePanelLayout). Verifier-caught satellites also removed: the `panels-vendor` manualChunks branch in `vite.config.js`, the orphaned `--header-height` alias in `tokens.css`.
- **Flag-OFF pin tests deleted** (`ShotListPage.desktopThreePanel`, `ShotDetailPage`, `ShotDetailPage.desktop`, ThreePanel tests) — the rollback path they protected dies here. Flag-ON tests converted by stripping now-meaningless flag mocks; unified-editor coverage fully retained (verified: every deleted file mocked `featureUnifiedShotEditor:false`).
- **No localStorage migration code** for the dead `sb:three-panel:*` keys (harmless orphans, by design).

## Convergence items (pinned at #433 review for 5c)

[...]

PR #437, opened 2026-06-10
fix(app): auto-reload once on stale lazy-chunk load failure

ted-design:
## Summary

Tiny hardening prompted by today's prod Sentry event (`SyntaxError: Unexpected token '<'` at `/library/talent`, 2:12pm EDT): a tab opened before one of today's three deploys requested a lazy route chunk whose hashed filename no longer existed; Firebase's SPA rewrite served `index.html`, which the browser tried to parse as JS. `RouteBoundary` contained it and reported to Sentry (`handled=yes`), but the user had to reload manually.

This PR makes that recovery automatic: a `vite:preloadError` listener (installed in `main.tsx` before render) reloads the page **once** to pick up the new build.

## Loop safety

A `sb:chunk-reload-at` sessionStorage timestamp guards a 60s window: if a preload error fires again within it (i.e. the deploy itself is broken, not stale), the handler does nothing — the error propagates exactly as today to `ErrorBoundary`'s chunk-aware "Reload required" fallback and Sentry. Per-tab by design (each stale tab gets one silent recovery). If sessionStorage is unavailable, it still reloads and the boundary remains the backstop.

## Scope

+134 lines, 3 files: `shared/lib/chunkReload.ts`, its test, one import + call in `main.tsx`. No behavior change for fresh sessions; independent of PR #436 (branched off main).

## Test plan

- [x] 5 unit tests (`chunkReload.test.ts`): first-error reload + event swallow, in-window suppression with propagation, post-window re-arm, storage-unavailable fallback, listener registration — 5/5 green
- [x] lint 0 warnings · vite build clean
- [ ] CI green
- [ ] Ted merge gate

### DIFF
--- a/src-vnext/main.tsx
+++ b/src-vnext/main.tsx
@@ -4,6 +4,9 @@ import * as Sentry from "@sentry/react"
 import "../tokens.css"
 import "./index.css"
 import { App } from "@/app/App"
+import { installChunkReloadHandler } from "@/shared/lib/chunkReload"
+
+installChunkReloadHandler()
 
 interface BootError {
   readonly type?: string
--- a/src-vnext/shared/lib/chunkReload.test.ts
+++ b/src-vnext/shared/lib/chunkReload.test.ts
@@ -1,0 +1,91 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { handlePreloadError, installChunkReloadHandler } from "./chunkReload"
+
+const GUARD_KEY = "sb:chunk-reload-at"
+
+function makeEvent(): Event {
+  const event = new Event("vite:preloadError", { cancelable: true })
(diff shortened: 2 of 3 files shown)


ted-design:
Round-1 disposition (`4f860992`):

- **Issue 1 (guard-key drift):** fixed — `RELOAD_GUARD_KEY` exported and imported by the test.
- **Issue 2 (comment block):** fixed — JSDoc collapsed to one line per house style; the storage-catch WHY comment kept.
- **Minor nit (idempotent install):** declined — single call site (`main.tsx`), and vitest isolates module registries per test file, so accumulation can't occur in the current or planned test structure; module-level state isn't warranted for a hypothetical.

Tests 5/5, lint 0 after the push. Converged — no further commits planned.

PR #438, opened 2026-06-10
feat(shots): port [ / ] prev-next + in-editor quick-add to the unified editor

ted-design:
## Summary

The named follow-up to #436: ports the two affordances that died with ThreePanel into the unified editor, flag-free, per the locked decisions (port WITH list-ordering context · quick-add at the bottom of the right rail · stay on the current shot after create).

## How it works

**`[` / `]` prev-next with list-ordering context.** The detail route deliberately knows nothing about the list's sort/filter state — the list page owns it. On shot click, `ShotListPage` snapshots the visible `displayShots` order into a per-tab sessionStorage key (`sb:shots:nav-order:<clientId>:<projectId>`); the editor reads the snapshot per keypress and navigates to the neighbor id. Clamps at the ends, no wrap — exact ThreePanel parity. Navigation-only, so no role gate (also parity). Deep links and new tabs have no snapshot → the keys no-op. Plain `[`/`]` only — cmd+[ browser history is untouched (the shortcut hook requires exact modifier match).

**In-editor quick-add.** New `footer` slot on `ShotDetailSidebar` hosts `ShotDetailQuickAdd` (a thin wrapper over the surviving `ShotQuickAdd` — same numbering, same `N`-to-focus, same batch-paste). Gated by `canDoOperational` (`!roleResolving && canManageShots(role)`, the 5b pattern), which also means the project-shots subscription it needs for next-number computation only ever mounts for operational roles — viewers/warehouse never subscribe. Returns null while loading so numbering can't run against an empty snapshot. After create: toast, stay on the current shot (new shot appends to the project list end, same as the list page).

## Verification

- 76 tests green across the 4 touched/new test files: snapshot lib pins (round-trip, scoping, malformed payloads, storage-error swallow), editor `[`/`]` matrix (prev/next, clamp, no-snapshot + absent-shot no-ops), quick-add gating matrix (producer renders in-rail / viewer absent / resolving absent).
- Adversarial pre-PR review (code-reviewer agent): **PASS**, no blockers. Confirmed the `handleShotClick` relocation has all 4 consumers below the new definition, the quick-add role gate maps onto the hardened `/shots` CREATE arms (admin / global-producer / project producer-crew), and `useShots` mounts exactly once, gated. One pre-existing page-wide note (global-crew non-member sees write affordances that rules deny → permission-denied toast) is the same accepted property as the status keys — not introduced here.
- lint 0 warnings · vite build clean.

## Test plan

- [x] `shotListNavOrder.test.ts` 5/5 · `ShotDetailPageUnified.test.tsx` 24/24 · `ShotListPage.test.tsx` + `ShotQuickAdd.test.tsx` green
- [ ] CI ui-checks (job `test`) — the arbiter
- [ ] Optional :5174 eyeball (keys + rail quick-add on real-shaped data)
- [ ] Ted merge gate

### DIFF
--- a/src-vnext/features/shots/components/ShotDetailPageUnified.tsx
+++ b/src-vnext/features/shots/components/ShotDetailPageUnified.tsx
@@ -29,6 +29,8 @@ import { ShotReferenceLinksSection } from "@/features/shots/components/ShotRefer
 import { SceneContextBanner } from "@/features/shots/components/SceneContextBanner"
 import { SceneDetailSheet } from "@/features/shots/components/SceneDetailSheet"
 import { ShotDetailSidebar } from "@/features/shots/components/ShotDetailSidebar"
+import { ShotDetailQuickAdd } from "@/features/shots/components/ShotDetailQuickAdd"
+import { readShotListNavOrder } from "@/features/shots/lib/shotListNavOrder"
 import { updateShotWithVersion } from "@/features/shots/lib/updateShotWithVersion"
 import { formatDateOnly, parseDateOnly } from "@/features/shots/lib/dateOnly"
 import { useAuth } from "@/app/providers/AuthProvider"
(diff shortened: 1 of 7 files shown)


ted-design:
Round-1 disposition: LGTM, zero blocking. The named optional test gap (`]` from index 0) is already pinned — the clamp test uses a single-element snapshot, so the current shot IS index 0 and both `[` and `]` assert no navigation (`orderIds[-1]`/`orderIds[1]` both undefined → guard returns). No further commits. Merging on Ted's standing authorization (CI green + bot converged + pre-PR adversarial review PASS).

PR #440, opened 2026-06-11
feat(shots): immediate talent selection + popover gap-close (redesign Phase 5d)

ted-design:
## Summary

Phase 5d gap-close per `outputs/2026-06-10-phase5d-build-spec.md` (the 8-agent audit found both §5d popovers already shipped in #402/#405 — this PR closes the deltas). Built via a 14-agent scout→build→adversarial-verify workflow; all per-writer verifiers, the persona rules-trace, and the full-diff review passed.

### TalentPicker — immediate selection (locked Decision A: debounced-draft)
- The open popover's local draft is the single source of truth (never re-derived from the refreshing `selectedIds` prop — no flicker when a write round-trips).
- Per-toggle UI is instant; the Firestore write is trailing-debounced ~500 ms and coalesced to **one** `save({ talent: ids, talentIds: ids })` per burst → one version snapshot. Flushed on close (the `handleOpenChange` close-diff IS the flush) and on unmount via a latest-ref.
- **Dual-field lockstep is preserved as a hard contract** (`talentIds` feeds array-contains queries; call-sheet/export resolvers fall back `talentIds` → `talent`). The page lambda needed zero changes — `ShotDetailPageUnified.tsx` has **no diff**.
- Failed bursts revert to last-saved and surface exactly one toast.
- Fixes the in-place `draft.sort()` mutation (`[...draft].sort()`); persisted order stays sorted (call sheets consume it).

### Search-vs-collapsed-groups bug fix
While a query is active, all collapsed groups force-mount (so cmdk can match their items) and the "Show N…" disclosures hide; clearing re-collapses. Acceptance pinned: a collapsed shortlist talent is found with its status badge.

### ProductQuickViewPopover hardening
`collisionPadding={8}` + `hideWhenDetached` (popper props pass through the existing ui/popover wrapper — no new wrapper). Dead `needsLookup` ternary removed; the legacy colourId-as-skuId fallback and its 6 tests are untouched.

### Eyeball-caught fixes (the :5174 matrix earned its keep)
- **Version snapshots were silently failing for every meta-line save** (pre-existing): mapShot-shaped shots carry nested `undefined`s inside `looks`, and `addDoc` rejects any `undefined` — the best-effort catch swallowed it, so talent/date/location saves minted **zero** history entries. `stripUndefinedDeep` in `buildShotSnapshot` fixes it (plain objects cleaned, array slots nulled, Timestamp-likes pass through). Verified live: a picker burst now mints exactly one `talent, talentIds` snapshot.
- **Escape was navigating away**: the page binds `Escape → navigate(-1)` (`ShotDetailPageUnified.tsx:338`, ported from ThreePanel); with a popover open, Escape dismissed it AND left the page. Both popovers now `stopPropagation` in `onEscapeKeyDown`. Negative-tested (the new test fails without the fix).

### Bundled fixes (spec §5)
- `addTalentToProject` auto-repair is gated on the **global** admin/producer claim (the write hits org-scoped `/clients/{cid}/talent`, rule `firestore.rules:363-365`) plus a per-client/project denied-memo — a crew editor with an orphaned talent link can no longer fire repeating denied writes.
- castingBoard permission-denied `console.error` quieted via `quietErrorCodes` in `useFirestoreCollection`, mirroring `useFirestoreDoc`'s `MEMBER_DOC_OPTIONS`; the error state (and the picker's flat-list degrade) is unchanged and now test-pinned.
- Orphaned `ProductSummaryStrip` deleted (zero consumers).
- a11y: decorative checkboxes `aria-hidden`, items `aria-checked` (cmdk owns `aria-selected` — pinned in a test as the accepted deviation), disclosure buttons `aria-expanded`/`aria-controls`, empty Booked group guarded.

## Tests
First-ever TalentPicker suite (16: partition incl. `passed`→Other, booked-first order, assigned hoist, flat degrade pin, search-finds-collapsed + re-collapse, one-save-per-burst at the 499/500 ms boundary, close-flush no-duplicate, unmount flush, dual-field lockstep, legacy names, revert+single-toast, a11y, no-mutation, Escape containment) · shotVersioning (4) · projectAssetsWrites gate (19) · useFirestoreCollection quiet (5) · useCastingBoard (3) · ProductQuickViewPopover 6 pre-existing stay green. Lint 0/0, build clean.

## Adversarial verification
- Persona (a) global-crew, no members doc: picker mounts, flat ungrouped list, no badges; the shot write is rules-denied at both update arms → one toast + silent revert, no write lands, no console spam, the repair gate returns before any org-scoped write (no denied-write loop).
- Persona (b) global producer, non-member private project: same degrade class; the repair gate correctly ALLOWS (the talent rule is global `isProducer()`).
- `:5174` eyeball matrix (second dev server, throwaway origin + throwaway project/talent, all deleted after): 1280px leftward-fit ✓ · ~1024px single-column with adjacent remove buttons clear ✓ · 390px Radix side-flip ✓ · bottom-row edge-hugging via collisionPadding ✓ · hideWhenDetached on rail scroll (hides + reattaches) ✓ · grouped view/badges/hoist/search/immediate-save verified live against prod-shaped data.

`firestore.rules` · `storage.rules` · `rbac.ts` · `PublicShotSharePage` · `ShotDetailPageUnified` (zero diff — capability block `:185-222` untouched). Both popovers stay presentational (no role/device hooks inside). No rename (Decision C), no colorway-strip work (Decision B deferred).


ted-design:
**Disposition (convergence declared — round 1, no blocking issues):**

1. **Concurrent-burst race in `commitDraft`** (non-blocking, per review): acknowledged and independently flagged by the pre-PR adversarial verifier + persona trace. Requires a burst-A save still in-flight (>500ms latency) when burst B commits AND A then rejecting — the revert clobbers B's `lastSavedRef` but still yields at most one toast and a consistent server state on the next echo. Accepted under locked Decision A; documented here and in the PR body rather than pushing a comment-only commit (re-review churn). If it ever manifests, the fix is a generation counter invalidating stale reverts — banked for the 5e picker touch.
2. **`resetDeniedTalentLinkMemoForTests` `@internal` JSDoc** (nit): agreed in spirit; the export already carries the `ForTests` naming convention used elsewhere in the codebase. Will fold the annotation into the next PR that touches `projectAssetsWrites.ts` (5f warehouse work) rather than re-triggering CI for a doc-comment.

No further commits planned for this PR.

PR #441, opened 2026-06-11
feat(shots): Phase 5e-I — resolver affordances/chrome outputs + device-authority subsumption

ted-design:
## Summary

Phase 5e-I of the redesign (spec: `2026-06-11-phase5e-build-spec.md`, Ted-approved with decisions A–G locked). First of the 3-PR 5e partition — **zero behavior delta except two named changes**:

1. **`featureSurfaceResolver` default flips ON** — never-customized producers/admins on tablet/desktop get the **table** default instead of card (the accepted change named since the Phase-4 flag-removal checklist). URL params and stored explicit choices still win; crew/warehouse/viewer resolution is byte-identical (pinned).
2. **Export keys to desktop** — the 768–1023px tablet Export button no longer mounts. It was a dead-end: `canExport` was true at ≥768px but the export route's `RequireDesktop` requires ≥1024px, so tablet taps got a toast+redirect.

## What changed

- **`resolveSurface` gains `affordances` + `chrome` outputs** (cut from Phase 4 per no-stubs; landing now WITH their first consumers). Every field has a named consumer; 5e-I values are pure device derivations reproducing today's gates. 5e-II reshapes them for `surface==='shoot'` behind `featureShootSurface`.
- **New `useResolvedSurface()` hook** — the shared resolution call site for list + detail. Always-on by design (the resolver flag gates only the list's viewMode/groupKey substitution). The opaque `effectiveRole` input is the 5e-III View-as interposition seam (comment-marked, not built).
- **Every `!isMobile` capability term rewired through `affordances`**: ShotDetailPageUnified (canEdit / canManageLifecycle / canUploadShotImages / canExport / inline Share / the status-control render fork), ShotListPage (canBulkPull / canRepair / canExport / canManageLifecycle), ShotVersionHistorySection (the internal isMobile kill-switch). **The rbac/global-claim terms stay byte-identical at the consumers** — resolver output never acts as a role source (the Phase-4 security law).
- **`featureShootSurface` added** (default OFF, `VITE_SHOOT_SURFACE` enable-only) — gates all 5e behavior; the Shoot shell (5e-II) mounts only off resolved `surface==='shoot'`.
- **`ProductColorwayStrip` extracted** to its own file (behavior-preserving) — the 5e-II Shoot shell + 5f Review presentational seam. First tests added.
- Resolving-gap behavior preserved exactly (raw device fallbacks while affordances resolve — pinned; no flash-of-missing-Export on desktop, no flash-of-appearing write affordances).

`firestore.rules` · `storage.rules` · `rbac.ts` · `PublicShotSharePage` · `useShotListState`'s resolution path · the ShotListPage `:596` JSX fork (5e-II's) · toolbar props.

## Test plan

- Built by a 12-agent writer→adversarial-verifier workflow (every partition PASS; zero-delta traced per gate × role × device × resolving-gap).
- **788/788 tests green across 10 files** (run locally): flags, resolveSurface (legacy-oracle matrix untouched + new role×device affordances/chrome characterization matrix), useResolvedSurface (12 new), ShotDetailPageUnified 26 (incl. tablet-export-hidden + mobile tap-row pins), ShotListPage 28 (incl. desktop-export-no-flash pin), ShotVersionHistorySection, ProductColorwayStrip, resolver inertness pins (unmodified, green).
- `tests/role-matrix.spec.ts` assertion (a) re-pinned to the table default — CI ui-checks is the arbiter for the E2E lane.
- Lint clean on all changed files; production build clean.

### DIFF
--- a/src-vnext/features/shots/components/ProductColorwayStrip.tsx
+++ b/src-vnext/features/shots/components/ProductColorwayStrip.tsx
@@ -1,0 +1,90 @@
+// Product colorway strip — pure-text typographic centerpiece (read-only; writes stay in the right rail).
+// Extracted from ShotDetailPageUnified (Phase 5e-I): the Shoot shell (5e-II)
+// and the Review surface (5f) reuse it — this extraction is the 5e/5f
+// presentational seam. Behavior-preserving: markup, classes, and testids are
+// identical to the private original.
+import { SectionLabel } from "@/features/shots/components/ShotDetailShared"
+import type { ShotLook } from "@/shared/types"
+
+export function ProductColorwayStrip({
+  looks,
+  activeLookId,
+}: {
+  readonly looks: ReadonlyArray<ShotLook>
+  readonly activeLookId: string | null | undefined
+}) {
+  const sorted = [...looks].sort((a, b) => (a.order ?? 0) - (b.order ?? 0))
+  const productCount = sorted.reduce((acc, l) => acc + (l.products?.length ?? 0), 0)
+  const resolvedActiveId =
+    activeLookId && sorted.some((l) => l.id === activeLookId)
+      ? activeLookId
+      : sorted.length > 0
+        ? sorted[0]!.id
+        : null
+
+  return (
+    <section
+      className="border-t border-[var(--color-border)] pt-4"
+      data-testid="product-colorway-strip"
+    >
+      <div className="flex items-baseline gap-2">
+        <SectionLabel>Products</SectionLabel>
+        {sorted.length > 0 && (
+          <span className="text-2xs text-[var(--color-text-subtle)]">
+            {productCount} {productCount === 1 ? "item" : "items"} &middot; {sorted.length}{" "}
+            {sorted.length === 1 ? "look" : "looks"}
+          </span>
+        )}
+      </div>
+
+      {sorted.length === 0 ? (
+        <p className="mt-1.5 text-sm text-[var(--color-text-muted)]">
+          No products yet. Add a look in the rail.
+        </p>
+      ) : (
+        sorted.map((look) => (
+          <div key={look.id} className="mt-3 first-of-type:mt-2">
+            <p className="label-meta">
+              {look.label || "Look"}
+              {look.id === resolvedActiveId ? <> &middot; Active</> : null}
+            </p>
+            {(look.products ?? []).length === 0 ? (
+              <p className="py-1 text-sm text-[var(--color-text-muted)]">
+                No products in this look.
+              </p>
+            ) : (
+              (look.products ?? []).map((p, i) => (
+                <div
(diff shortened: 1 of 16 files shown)


ted-design:
## Review disposition (round 1)

**Item 1 (fallback asymmetry) — FIXED in `b17029b5`.** The three role-gated ShotListPage fallbacks now use `?? !isMobile`, the literal pre-change device expression, matching ShotDetailPageUnified. The difference was unreachable in practice (authLoading without roleResolving), but the zero-delta claim is now literal in both files.

**Item 2 (Chrome.toolbar unreachable variants) — DECLINED, documentation-of-intent.** The `'minimal'|'none'` union members encode the 5e-II contract from the approved spec (§The API: the Shoot list ships `toolbar:'minimal'`). The chrome SLOT has live consumers at 5e-I (`statusControl` drives the detail-page fork; `viewSwitcher`/`quickAdd` values are characterization-pinned); narrowing the union to `'full'` now would mean a type-churn PR at 5e-II for zero safety gain. The no-stubs rule targets dead output slots — the Phase-4 cut — not union members of a consumed field.

**Item 3 (no env-var rollback) — ACKNOWLEDGED, by precedent.** Enable-only env semantics are the established pattern from the `featureUnifiedShotEditor` flip (#434): rollback = flip the default back + redeploy (~15 min via deploy-live), same operational posture as every flag flip in this initiative. Noted as the operational reality in the spec's flag-strategy section.

**Item 4 (multi-line comment blocks) — DECLINED for the constraint-bearing blocks.** The security-law and resolving-gap comments state invariants the code cannot show (the exception the house style carves out); compressing them to one line would drop the constraint, not the noise. Agreed in principle for pure narration — none of the flagged blocks are.

**Item 5 (duplicated flag mock) — AGREED, deferred with a trigger.** Extracting a shared mock helper is warranted when a third fork-test file appears (likely 5e-II). Registered in the 5e-II scope notes.

Convergence declared — no further commits planned beyond CI on `b17029b5`.

ted-design:
## Review disposition (round 2 — final)

**Item 1 (double `?? []` in ProductColorwayStrip) — DEFERRED to 5e-II.** That shape is the pre-existing code, preserved deliberately: the extraction contract was byte-faithful/behavior-preserving, and the hoist — while behavior-identical — belongs with the first real change to the file. 5e-II consumes the strip in the Shoot shell and takes the one-liner there (registered in its scope notes alongside the round-1 item 5 mock-helper trigger).

**Item 2 (extra `useIsTablet` mock entry) — ACKNOWLEDGED, defensive padding.** `useResolvedSurface` derives the device from `useIsMobile` + `useIsDesktop` only (matching ShotListPage's derivation). The mock entry is inert; 5e-II's test-mock consolidation (round-1 item 5) removes it.

Both items are cosmetic on a green, converged PR — holding the no-churn line. **Convergence final at `b17029b5`.**

PR #442, opened 2026-06-11
feat(shots): Phase 5e-II — the Shoot surface shell, mobile capability flips, offline durability, invite hardening

ted-design:
## Summary

Phase 5e-II per the approved spec (`2026-06-11-phase5e-build-spec.md`, decisions A–G locked). Crew get a purpose-built compact on-set surface behind the new `featureShootSurface` flag (default OFF — **zero user-visible change at merge** except the three named flag-independent fixes below).

## What ships

**The Shoot shell** (resolved `surface==='shoot'` + flag, at the existing routes):
- **Detail**: explicit Back → the shots list (never history-back) · shot-identity header · read-only hero (Decision B) · comments · FULL scene direction (the old banner truncated to 100 chars and was invisible below 640px) · read-only product strip · prev/next tap targets · **sticky bottom 44px status tap-row** (Decision A) with safe-area padding. Planning editors NOT MOUNTED.
- **List**: compact tappable rows in display order (Decision G), minimal chrome; **legacy `projectId==''` shots filtered for crew** with a quiet note and excluded from the nav order (Decision D — they're rules-uneditable for crew); legacy deep-links render read-only.
- Surface-keyed, not device-keyed: tablet/desktop crew get the same shell, centered (Decision F).
- Pending/offline indicators on the tap-row + composer; the composer queues offline with a quiet note.

**Mobile capability flips** (same flag): the `!isMobile` device terms drop for plan-build roles via the resolver's flag-aware affordances — full-matrix invariance test pins `surface/viewMode/groupKey/viewSource` flag-independent and flag-off values byte-identical to 5e-I.

**Offline durability** (Decision C, own commit `2cecd071`): Firestore `persistentLocalCache(multi-tab)` — offline writes queue durably in IndexedDB and replay on reconnect (previously RAM-only, silently lost on reload — the audit's headline on-set gap). **App-wide by nature** (cannot be flag-forked); revert that one commit to remove the blast radius. No-IndexedDB environments fall back to memory cache (= today). Emulator wiring unchanged.

**Invite hardening** (Decision E): `setUserClaims` CF writes project member docs atomically with the claims update (the existing-user lane previously stranded crew with claims but no membership = UI-enabled, write-dead). Legacy role 'editor' rejected for NEW invites only.
⚠️ **`functions/index.js` changed → requires a manual Ted-gated `firebase deploy --only functions` after merge** (deploy-live is hosting-only). Until then the client is fully functional against the old CF (the idempotent client-side member write is kept, removal trigger documented).

**Three flag-independent rules-honesty fixes** (spec rules-vs-UI matrix items 3+5 + banked):
1. FAB `?create=1` consume path now respects `showCreate` (mobile viewers no longer surface the create dialog; waits out role resolution so producer intent is never swallowed — pinned).
2. Lifecycle **Transfer/Copy hidden without the GLOBAL producer/admin claim** the move rule requires (project-promoted producers saw rules-denied items). Archive/duplicate/delete stay effective-role.
3. **TalentPicker stale-settlement race** (PR #440 banked): per-commit generation counter; both new tests verified RED pre-fix. Plus the missing LocationPicker Escape guard (negative-tested) and the #441 strip cleanup.

## E2E + rules lanes

- ui-checks gains a **second flag-ON build** (`VITE_SHOOT_SURFACE=1`) + targeted `tests/shoot-shell.spec.ts` pass (member-crew phone day-loop: shell renders, planning editors not mounted, **status tap writes AND the version snapshot doc lands** — the 5d-bug-shape regression pin; legacy deep-link read-only). The default lane runs grep-inverted and stays the flag-OFF contract. Lane gate documented in QUARANTINE.md with owner+phase (not a quarantine).
- `shots.rules.test.ts` 39→44: member-crew update→versions-create sequence, non-member denied on both, comment-create pins (incl. the NAMED rules-wider-than-UI viewer allowance — 5f Q4, deliberately untouched).
- The member-crew fixture already existed from 5b; the load-bearing NON-member crew fixture untouched.

## Verification

- Built by an 18-agent scout→write→adversarial-verify workflow (4 scouts + 7 writer/verifier pairs, ALL PASS; one verifier independently neutralized both small fixes to prove the new tests are the sole pins).
- Gates run by the orchestrator: **916 unit tests green** across all touched files, lint clean, build clean, `node --check` on functions.
- **Live :5174 eyeball at phone width** (throwaway project, deleted after; Q2-26 No. 3 never opened): shell blocks in order ✓ sticky tap-row ✓ status tap persisted through the hardened rules ✓ prev/next + counter ✓ Back→list ✓ offline composer note ✓ producer view unchanged ✓. **Two real findings fixed + pinned** (`949cbc85`): the FAB overlapped the sticky status bar; producer rail copy leaked into the shell's empty product state.
- Honest residual: true airplane-mode queue-replay semantics have no unit oracle — covered by Firestore's standard offline mode + the CI flag-ON lane asserting the snapshot write; the simulated-offline eyeball verified the indicator copy only.

`firestore.rules` · `storage.rules` · `rbac.ts` · `PublicShotSharePage` · the load-bearing non-member crew E2E fixture · `role-matrix.spec.ts` · `global.setup.ts`.


ted-design:
## Disposition — Codex P2 + CI round 1

**Codex P2 (filtered shell list) — FIXED in `063d76f6`, rule-grounded and accepted.** The shell renders no filter controls, so a lingering deep-link `?status=…`/`?q=…` silently subset the on-set list — contradicting locked Decision G (full project list). `useShotListState` gains an additive `unfilteredSortedShots` (alive + display sort order, no condition/query filtering); the shell list, legacy-hidden count, and the prev/next nav-order snapshot all derive from it. Pinned with a real serialized `?filters` param and RED-verified against the pre-fix derivation. Producer surfaces untouched.

**CI round-1 vitest/build failures — FIXED in `6a1c4189`.** A pre-existing component-adjacent `FloatingActionBar.test.tsx` (not under `__tests__/`, missed by the targeted local runs) broke when the FAB gained its `useResolvedSurface` call for the Shoot-shell suppression (itself a :5174 eyeball finding). The five existing route/device pins now run with steerable surface/flag mocks — behaviors unchanged — and the three suppression pins live alongside them; the duplicate file is deleted. 8/8 green.

ted-design:
## Disposition (round 2 — convergence)

**Item 3 (overly broad init catch) — FIXED in `8a19ea68`, rule-grounded.** The known already-initialized (HMR) case falls back to `getFirestore`; everything else (IndexedDB schema errors, SDK mismatches) rethrows instead of silently degrading the app-wide cache.

**Item 2 (FAB test timing) — FIXED in `8a19ea68` for the two positive pins** (`waitFor` on the param-clear / dialog-open conditions). The third pin asserts a negative steady-state ("the effect declined to fire while the role resolves") — there is no condition for `waitFor` to await; its bounded ticks stay with the rationale documented in-code.

**Item 1 (useShotPendingWrites path construction) — DECLINED as churn.** The bot's own framing: "not a bug, just a readability improvement." Folded into the next file-touching change.

**Item 4 (FAB hook cost) — DECLINED; the suggestion is structurally impossible.** Hooks cannot move below an early return (Rules of Hooks); `useResolvedSurface` is memoized, subscription-light, and shared app-wide. Cost accepted as negligible.

**Item 5 (chunk `writeProjectMemberDocs`) — DEFERRED with the inherited pattern.** `assignToProjects` is a hand-picked project list (single digits in practice); chunking both lanes belongs to the 5f rules/functions pass if member-doc writes ever batch at scale.

**Convergence declared at `8a19ea68`** — no further commits planned beyond CI.

PR #444, opened 2026-06-12
fix(products): product merge crash on a SKU with no name (#7)

ted-design:
**Phase B (bug 1 of 2)** of the talent/casting/products/shots fix batch.

## The crash
The Product Merge Wizard failed with:
> `[executeProductMerge] Step 1 (transferNewSkus) failed: WriteBatch.set() called with invalid data. Unsupported field value: undefined (found in field name in document clients/unbound-merino/productFamilies/…/skus/…)`

whenever the loser product had a "new" (unmatched) SKU **missing a `name`** — legacy/imported SKUs can lack it even though `name` is the only required SKU type field. Step 1 wrote `name: sku.name` directly and, unlike Steps 2–5, skipped `sanitizeForFirestore()`.

## Fix
- Default the written name to `sku.name ?? colorName ?? skuCode ?? "Untitled"` — a nameless SKU still merges with a human-readable label, never `undefined`.
- Wrap the Step 1 payload in `sanitizeForFirestore()` (parity with Steps 2–5); also strips nested `undefined` (e.g. `assetRequirements`).
- Defensively sanitize **Step 6** (`updatePullReferences`), where a re-pointed pull item could inject an `undefined` `familyName`.

## Verification
- New `productMergeWrites.test.ts` (**5 tests**): a production-shaped repro of the missing-`name` SKU (asserts no `undefined` in the batch payload + a defined name), the `name → skuCode → "Untitled"` fallback chain, nested-undefined stripping, and the happy path / id-map.
- Exposed `transferNewSkus` via an `@internal` `productMergeWritesForTests` export (matches the repo's `…ForTests` convention).
- **Lint clean · 5 tests pass · production build succeeds.**

No rules/functions changes. CI `ui-checks` (job `test`) is the arbiter.

### DIFF
--- a/AI_RULES.md
+++ b/AI_RULES.md
@@ -135,6 +135,16 @@ When creating or updating tags anywhere in the codebase:
 - **Export builder:** Inline on block data (persisted with export document)
 - **Reusable component:** `ColumnSettingsList.tsx` — DnD column list shared by shot table popover and share dialog
 
+### Scene Infrastructure Pattern (Sprint S29)
+
+- **Numbering:** Shots in scenes use letter suffixes: `{sceneNumber}{A-Z/AA-ZZ}`. Use `formatSceneShotNumber(sceneNumber, index)` — never construct manually. `parseSceneShotNumber("51A")` → `{ base: 51, suffix: "A" }`. Ungrouped shots in scene mode use raw `String(n)` (no `formatShotNumber` padding). Never use `Number()` or `padStart()` on shot numbers — use `localeCompare({ numeric: true })` for sorting.
+- **Scene colors:** Import from `lib/sceneColors.ts` (not from `SceneHeader.tsx`). `getSceneColor(key)` resolves to hex or falls back to `var(--color-text-subtle)`.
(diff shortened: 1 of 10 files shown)


ted-design:
Addressed review notes in `3b45d6c`:
- **🔴 winnerName:** `familyName` now falls back to the re-pointed item's own name (`winnerName ?? item.familyName`) so an unnamed winner can't silently drop it; kept the defensive sanitize. Applied to all three family-repoint sites for consistency.
- **Helper rename:** `undefinedFields` → `topLevelUndefinedFields`.
Skipped with rationale: a dedicated `updatePullReferences` integration test would need disproportionate multi-project `getDocs` mocking for a now-defensive line; the Step-1 repro covers the undefined-rejection class. Convergence.

ted-design:
Follow-up `a25116c1`: reworded the inline comment to explain the WHY (legacy families can lack the typed-required `styleName` at runtime — the same class of bug this PR fixes), which is also why the `winnerName ?? …` guard is intentional rather than dead code (dead per the type-checker, live per production data). Step-6 left untested by design (acknowledged — disproportionate multi-project getDocs mocking for a defensive line). Convergence.

PR #445, opened 2026-06-12
fix(talent): replaced headshot keeps coming back — stale image cache (#2)

ted-design:
**Phase B (bug 2 of 2)** of the talent/casting/products/shots fix batch.

## The bug
Replacing a talent's profile image didn't stick — the **old image kept reappearing** after remove + replace (reported on **Jesse Dunphy**, who showed a wrong product/grey-tee shot).

## Diagnosis (against live Firestore + Storage)
The file at Jesse's path was **already correct** (a B&W portrait he'd uploaded). Verified live and confirmed by Ted. The app was serving a **stale cached image**:
- Headshots uploaded to a **fixed** path `images/talent/<id>/headshot.webp`, so every replacement overwrote the **same** Storage object and reused the **same** download URL.
- The browser HTTP cache **and** the in-memory `resolveStoragePath` cache (30-min TTL, keyed by path) both kept serving the old bytes.
- Gallery/casting images already avoid this by using unique filenames (`crypto.randomUUID()`).

## Fix
- `headshotStoragePath()` → a **unique** filename per upload (`headshot-<uuid>.webp`), used by `createTalent` + `setTalentHeadshot`. A unique path = a fresh URL no cache layer can stale-serve; the previous object is deleted afterwards (`previousPath` cleanup now always fires on replace).
- New `invalidateStoragePath(path)` drops a path from the resolve cache; called on replace/remove (belt-and-suspenders).

## Verification
- **7 new tests** (`talentWrites.test.ts` + `resolveStoragePath.test.ts`): unique-path regression, no fixed name survives, two replacements differ, `createTalent` headshot path, all three image fields nulled on remove, cache-invalidation forces a refetch.
- **Lint clean · 7 tests pass · production build succeeds.**
- **Live E2E** on the dev server: a fresh load now renders Jesse's correct portrait; Ted confirmed it's the right image.
- **Adversarial code-review: APPROVE** — call sites pass `previousPath` (no Storage orphan leak); display resolves the new path everywhere.

Note: existing records (like Jesse) still carry the old fixed `headshot.webp` path with correct bytes — a hard-refresh or the next re-upload (now unique) clears any lingering client cache. No data migration required.

### DIFF
--- a/src-vnext/features/library/lib/talentWrites.test.ts
+++ b/src-vnext/features/library/lib/talentWrites.test.ts
@@ -97,7 +97,7 @@ describe("talent headshot writes (replaced-image fix)", () => {
     expect(deleteObject).not.toHaveBeenCalled()
   })
 
-  it("removeTalentHeadshot clears all three image fields and invalidates the cache", async () => {
+  it("removeTalentHeadshot clears all three image fields + invalidates cache (no object delete — symmetric with replace)", async () => {
     const prev = "images/talent/t1/headshot-xyz.webp"
     await removeTalentHeadshot({ clientId: "c", userId: "u", talentId: "t1", previousPath: prev })
 
@@ -106,6 +106,6 @@ describe("talent headshot writes (replaced-image fix)", () => {
     expect(written.headshotUrl).toBeNull()
     expect(written.imageUrl).toBeNull()
     expect(invalidateStoragePath).toHaveBeenCalledWith(prev)
-    expect(deleteObject).toHaveBeenCalledTimes(1)
+    expect(deleteObject).not.toHaveBeenCalled()
   })
 })
--- a/src-vnext/features/library/lib/talentWrites.test.ts
+++ b/src-vnext/features/library/lib/talentWrites.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect, vi, beforeEach } from "vitest"
+
+const updateDoc = vi.fn()
+const addDoc = vi.fn()
+const uploadBytes = vi.fn()
+const getDownloadURL = vi.fn()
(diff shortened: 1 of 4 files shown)


ted-design:
Addressed review notes in `58fd9ad`:
- **Codex P2 (casting-share headshots):** confirmed shares **snapshot** the headshot download URL (`resolveTalentForCastingShare`), so I now **do not delete** the old object on replace — only invalidate the cache. The unique-filename change alone fixes the stale display. Explicit `removeTalentHeadshot` still deletes (deliberate action). This also moots the orphan-on-replace concern. Test updated to assert no deletion on replace.
- **Comments:** trimmed to one line. **Test isolation:** clear the module cache in `beforeEach`.
Skipped: `deleteTalent` (deprecated, soft-delete doesn't touch Storage) cache-invalidation — low value. Convergence.

ted-design:
Follow-up `04cd39f5`: corrected the misleading comment — the old object is **intentionally left orphaned** on replace (accepted storage tradeoff) to keep published casting shares working; it is NOT swept on hard-delete. Dropped the redundant `?? null`. The in-flight pending.delete race is a pre-existing cache-module limitation (acknowledged, not introduced here). Convergence.

ted-design:
Follow-up `fc06e6da`: resolved the remove-vs-replace asymmetry — `removeTalentHeadshot` now also leaves the old object (just clears the doc fields + cache), so an accidental remove can't break a casting share that snapshotted it. Symmetric + share-safe. Added the `setTalentHeadshot` return-type annotation. The cast-type note (sanitize return) matches the existing Steps 2–5 pattern — non-blocking. Convergence.

PR #447, opened 2026-06-12
feat(casting): per-image + per-folder casting-board image visibility (#5)

ted-design:
## What & why
The admin casting board showed only the headshot even when a talent had gallery + casting-session images — the **public** share already rendered them, but producers had no way to surface or curate them on the board. This adds **per-image AND per-folder (casting session) visibility** controlling which of a talent's existing images appear on their casting-board card. (Issue #5 from the 2026-06-12 talent/products/shots fix batch.)

## Approach
- **No new subcollection.** The casting board entry doc's id *is* the `talentId`, so visibility lives as `hiddenImageIds[]` / `hiddenSessionIds[]` fields on the existing `castingBoard/{talentId}` doc. Both image arrays and casting sessions carry stable `id`s.
- **No `firestore.rules` change.** The project-scoped wildcard (`firestore.rules:921-928`) already gates `castingBoard` writes to admin / producer / warehouse (crew read-only) — matching the existing `canEdit = canManageCasting(role) && !isMobile` UI gate. Rules-traced for admin / global-producer / crew.
- New **pure lib** `castingBoardImages.ts` (`buildBoardImageModel` + `resolveVisibleBoardImages`) is the single source of truth for "what's visible," shared by the card strip, the admin sheet, and the share resolver.
- **`AdminTalentDetailSheet`** now renders the gallery + one section per casting-session folder, with per-image Eye/EyeOff and per-folder hide toggles (gated on `canEdit`). Also fixes a **latent bug**: the old Portfolio grid read `img.url`, but the field is `downloadURL`/`path` (resolved via `useStorageUrl`) — which is why images never rendered.
- **`CastingCard`** shows a compact visible-images strip (max 4 + `+N` overflow), zero-regression when there are no extras.
- **`resolveTalentForShare`** excludes hidden images/sessions so curation is respected in public shares too — all `MAX_*` caps + the `portfolio` visibility gate untouched.
- `updateCastingEntryVisibility` writes via `sanitizeForFirestore`.

## How it was built
Scout → build → adversarial-verify dynamic workflow (10 agents): 2 scouts, 3 build slices (data/lib/write · admin sheet · card + share), 5 adversarial verifiers (per-slice + rules-trace + full-diff). All 5 verifiers PASS; only low-severity cosmetic notes (a pre-existing Radix `DialogContent` missing-description warning).

## Tests / gates (verified locally)
- New: `castingBoardImages` (13), `resolveTalentForShare` (4), `castingWrites` (2), `CastingCard` (6), `AdminTalentDetailSheet` (5); `useCastingBoard` mapper (+ focused case).
- CI-adjacent still green: `CastingBoardPage` (7), `TalentPicker` (18).
- **59 tests green** total run locally · `npm run build` clean (exit 0) · eslint exit 0 · no new tsc errors in touched files.

## Notes
- Behavior-free `data-testid` only; immutability preserved (toggles recompute new arrays). Scope = casting feature + a 4-line additive `CastingBoardEntry` type change.
- ⚠️ Node 24 on Actions runners since 2026-06-16 — if CI breaks, suspect checkout@v4/setup-node@v4/auth@v2 first.


ted-design:
Thanks both — dispositions for the feedback, all addressed in `6a506ff0`:

**Codex P2 — folder-inherited hidden state (real bug, fixed).** `buildBoardImageModel` now tracks `individuallyHidden` vs `folderHidden` separately. The per-image toggle is **disabled** when an image is hidden only via its folder, and `toggleImage` early-returns — so it can no longer append the id to `hiddenImageIds` without revealing the image (and strand it after the folder is shown again). New regression test asserts the inert toggle fires no write; the lib test asserts a both-hidden image stays individually hidden after un-hiding the folder.

**claude #1 — Rule #4 (fixed).** Added an Approved-exception entry in `CLAUDE.md §4` for `hiddenImageIds`/`hiddenSessionIds` with the no-rules-change rationale (the project-scoped wildcard already gates `castingBoard` writes to admin/producer/warehouse, matching the `canManageCasting` UI gate).

**claude #2 — stale-read toggle race (fixed).** The sheet now holds optimistic `pendingHidden` state and computes the next array from it, so two toggles inside one Firestore round-trip compose instead of dropping one (also gives instant feedback). Resets on talent switch / sheet reopen. New test composes two rapid hides and asserts the second write carries both ids.

**claude #4 — `hasHiddenCandidates` JSDoc (fixed).** Clarified to "precondition flag, not a count."

**claude #5 — `Set` in the share resolver (fixed).** Sets built once per talent; filtering + all `MAX_*` caps + the `portfolio` gate unchanged.

**claude #3 — overflow-chip tokens (addressed).** No semantic overlay token exists, and the chip is a scrim over an arbitrary photo where a theme token would be wrong; aligned it to the card's own in-file overlay convention (`bg-black/45`, matching the checkbox/menu at lines 94/107).

62 casting + TalentPicker tests green, build + lint clean, and an independent code-reviewer pass returned GO on the P2 + optimistic-state changes.

PR #448, opened 2026-06-12
feat(shots): merge two shots into one via a ShotMergeWizard (#6)

ted-design:
## What & why
Lets a user combine two shots into one on the Shots list — Ted's example: merge "Men's Linen Beach Shorts (Denim)" + "Men's Tropical Linen Shirt (White)" into a single styled shot. Mirrors the existing **Product Merge Wizard** so it feels native. (Issue #6 from the 2026-06-12 batch; built to the approved `outputs/2026-06-12-shot-merge-spec.md` + engineering contract.)

## UX
- Select **exactly 2** shots → a **Merge** action in `BulkActionBar` (gated `canManageShots`, enabled only at 2 selected) → a 3-step **`ShotMergeWizard`**: **Compare** (pick primary, swap, choose mode) → **Preview** (counts + soft-delete warning + confirm checkbox) → **Result** (summary + View shot).

## Merge semantics
- **Mode A — Combine into one outfit look** (default): all products from both shots union into the primary's active look (dedup by `familyId::skuId::colourId`), producing one look. **Mode B — Keep as separate looks**: both shots' looks are kept (id-collision remap + re-ordered).
- Union **talent/talentIds** (lockstep), **referenceLinks** (by url), **tags** (by id). Primary's **title/date/location/status/shotNumber/laneId/sortOrder** win. `heroImage` = primary ?? secondary. Notes concatenate into **`notesAddendum`** with a `— merged from "<title>" —` divider.
- Loser is **soft-deleted** (`deleted:true, deletedAt`), recoverable.

## Key engineering decisions (confirmed in code)
- **`looks[]` is the source of truth**; root `products[]` is a legacy fallback → the merge keeps looks canonical and **mirrors root `products` to the active look**. Legacy root-products-only shots are normalized into a synthesized look first.
- **Deep-sanitize the patch** (`sanitizeForFirestore`): `buildShotWritePayload` only strips *top-level* undefined, but `looks[]` carry nested undefined that Firestore would reject (same class as the prior version-snapshot bug).
- **Never write `notes`** — it's a `BLOCKED_FIELD`; concatenation targets `notesAddendum`.
- `executeShotMerge` writes the **primary first** (with a version snapshot), then soft-deletes the loser (own snapshot) — a failed primary write never orphans a deleted secondary.
- The planner (`buildShotMergePlan`) is **pure** (no Firestore) → fully unit-tested.

## How it was built
Scout → build → adversarial-verify dynamic workflow (9 agents): 2 scouts, 3 build slices (planner+writer · wizard · bar+host), 4 adversarial verifiers (planner / wizard+host / rules-trace / full-diff). All 4 PASS; only low cosmetic notes (one of which — a "Talent unioned" label that rendered the *added* count — is fixed in this PR).

## Tests / gates (verified locally)
- `shotMergeWrites` (19), `ShotMergeWizard` (6), `BulkActionBar` (6); CI-adjacent `ShotListPage.test.tsx` + `ShotListPage.unifiedEditorFork.test.tsx` (33) — **64 green** total · `npm run build` clean · eslint exit 0 · no new tsc errors.
- **Rules-traced**: no `firestore.rules` change — the shots update arm already governs both writes for admin/producer/crew; UI gate (`canManageShots`) matches.

## Notes
- v1 is **pairwise** (exactly 2); N-way deferred. Immutability preserved (planner builds new objects). Behavior-free `data-testid` only.
- ⚠️ Node 24 on Actions runners since 2026-06-16 — if CI breaks, suspect checkout@v4/setup-node@v4/auth@v2 first.

### DIFF
--- a/src-vnext/features/shots/components/BulkActionBar.tsx
+++ b/src-vnext/features/shots/components/BulkActionBar.tsx
@@ -1,5 +1,5 @@
 import { useMemo, useState } from "react"
-import { Layers, MapPin, Tag, Trash2, User } from "lucide-react"
+import { GitMerge, Layers, MapPin, Tag, Trash2, User } from "lucide-react"
 import { toast } from "sonner"
 import { Button } from "@/ui/button"
 import { Checkbox } from "@/ui/checkbox"
@@ -46,6 +46,7 @@ interface BulkActionBarProps {
   readonly onBulkDeleteOpen: () => void
   readonly onClearSelection: () => void
   readonly onGroupSceneOpen?: () => void
+  readonly onMergeOpen?: () => void
   // Flags
   readonly canShare: boolean
   readonly canExport: boolean
@@ -68,6 +69,7 @@ export function BulkActionBar({
   onBulkDeleteOpen,
   onClearSelection,
   onGroupSceneOpen,
+  onMergeOpen,
   canShare,
   canExport,
   locations,
@@ -305,6 +307,17 @@ export function BulkActionBar({
           >
             <Layers className="mr-1.5 h-3.5 w-3.5" />
             Set Scene
+          </Button>
+        )}
+        {onMergeOpen && canManageShots(role) && (
+          <Button
+            variant="outline"
+            size="sm"
+            disabled={selectedIds.size !== 2}
+            onClick={onMergeOpen}
+          >
+            <GitMerge className="mr-1.5 h-3.5 w-3.5" />
+            Merge
           </Button>
         )}
         {canManageShots(role) && (
--- a/src-vnext/features/shots/components/ShotListPage.tsx
+++ b/src-vnext/features/shots/components/ShotListPage.tsx
@@ -47,6 +47,7 @@ import { useLanes } from "@/features/shots/hooks/useLanes"
 import { SceneHeader } from "@/features/shots/components/SceneHeader"
 import { SceneDetailSheet } from "@/features/shots/components/SceneDetailSheet"
 import { GroupIntoSceneDialog } from "@/features/shots/components/GroupIntoSceneDialog"
+import { ShotMergeWizard } from "@/features/shots/components/ShotMergeWizard"
 import { createLane, assignShotsToLane, ungroupAllShotsFromLane, deleteLane } from "@/features/shots/lib/laneActions"
 import { writeShotListNavOrder } from "@/features/shots/lib/shotListNavOrder"
 import { Skeleton } from "@/ui/skeleton"
@@ -91,6 +92,7 @@ export default function ShotListPage() {
   const [bulkDeleteOpen, setBulkDeleteOpen] = useState(false)
   const [renumberOpen, setRenumberOpen] = useState(false)
   const [groupSceneOpen, setGroupSceneOpen] = useState(false)
+  const [mergeOpen, setMergeOpen] = useState(false)
   const [deleteSceneTarget, setDeleteSceneTarget] = useState<{ id: string; name: string } | null>(null)
   const [editSceneId, setEditSceneId] = useState<string | null>(null)
   const [collapsedScenes, setCollapsedScenes] = useState<ReadonlySet<string>>(new Set())
@@ -469,6 +471,7 @@ export default function ShotListPage() {
           role={role}
           onShareOpen={() => setShareOpen(true)}
           onGroupSceneOpen={() => setGroupSceneOpen(true)}
+          onMergeOpen={() => setMergeOpen(true)}
           onExportClick={() => navigate(`/projects/${projectId}/export?preset=shot-list`)}
           onCreatePullOpen={() => setCreatePullOpen(true)}
           onBulkDeleteOpen={() => setBulkDeleteOpen(true)}
@@ -977,6 +980,19 @@ export default function ShotListPage() {
         user={user}
         onDeleted={clearSelection}
       />
+
+      {selectedShots[0] && selectedShots[1] && selectedShots.length === 2 && (
+        <ShotMergeWizard
+          open={mergeOpen}
+          onOpenChange={setMergeOpen}
+          clientId={clientId}
+          user={user}
+          shotA={selectedShots[0]}
+          shotB={selectedShots[1]}
+          projectId={projectId}
+          onMerged={clearSelection}
+        />
+      )}
 
       <RenumberShotsDialog
         open={renumberOpen}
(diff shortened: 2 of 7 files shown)


ted-design:
Thanks both — dispositions, all addressed in `0759f161`:

**Codex P2 + claude Medium — wizard unmounts mid-flow (real bug, fixed).** You're right that the *success path itself* triggers it: soft-deleting the loser drops it from `displayShots` → `selectedShots` falls to 1 → the `selectedShots.length === 2` host guard goes false and the dialog unmounts before the Result step, leaving `mergeOpen` true. `ShotListPage` now snapshots the pair into `mergeShots` state at open time (`openMerge`) and hosts the wizard on that frozen array, clearing it on close. The wizard is now immune to selection changes while open.

**claude Critical — status labels (fixed).** Replaced the local `STATUS_LABELS` ("To do"/"Complete", which CLAUDE.md prohibits) with `getShotStatusLabel` from `statusMappings` → canonical Draft / In Progress / Shot / On Hold.

**claude low — empty arrays written to absent fields (fixed).** `buildShotMergePlan` now includes `talent`/`talentIds`/`referenceLinks`/`tags` only when non-empty. The union is always a superset of the primary's, so omitting an empty result never drops existing data. +2 tests lock this.

**claude low — non-atomic soft-delete (documented).** Added a `TODO(v2)` on `executeShotMerge` to batch the two writes once version snapshots support it; sequential-with-primary-first (no orphan on primary failure) is the accepted v1 posture.

**claude low — `referenceLinks` dedup keeps primary's title on URL collision.** Intentional "primary wins" semantic; leaving as-is.

**claude note — test gap for the unmount race.** The fix is structural — the frozen `mergeShots` snapshot makes the race impossible by construction (the wizard no longer reads `selectedShots`), which is exactly the snapshot fix you both proposed. A full `ShotListPage` selection→merge→loser-removed integration test is high-setup for low marginal value over the structural guarantee, so I've covered the cheap/high-value behavior (empty-array guard) with unit tests and will confirm the merge flow in the pre-merge eyeball. Happy to add the integration test if you'd prefer it gated.

66 tests green (shotMergeWrites 21 / wizard 6 / bar 6 / ShotListPage host 33), build + lint clean.

ted-design:
Round 2 dispositions — `355f82d1`:

**🔴 Rule 6b — non-atomic write.** Fixed rather than deferred: `executeShotMerge` now applies the merge as a **single Firestore `WriteBatch`** — the primary's merged patch and the secondary's soft-delete commit together or not at all, so a partial failure can never leave a half-merged pair. (Version snapshots remain best-effort audit writes fired after the atomic commit — they can't roll back the merge, which is the right posture for audit.) Orchestration tests rewritten to assert one batch / commit-once / atomic-commit-failure-mutates-nothing / clientId guard. This removes the data-consistency hole entirely — no user-approved limitation needed.

**🟡 Design system — card title.** Now uses `heading-subsection` instead of raw `text-sm font-semibold`.

**🟡 A11y — mode selector.** Switched from `aria-pressed` toggle buttons to `role="radiogroup"` + `role="radio"`/`aria-checked` (exactly-one-selected semantics).

**🟢 `loading="lazy"`** added to the compare-card hero images.

**🟢 `openMerge` guard.** Reordered so `selectedShots.length === 2` is the primary gate; the `a && b` destructure stays purely for TS tuple-narrowing (TS can't narrow a tuple from `.length`), with a comment to that effect.

67 merge-suite tests green (incl. new atomic-batch orchestration cases), ShotListPage host tests green, build + lint clean. An independent code-reviewer verified the atomicity refactor (GO; tests non-vacuous).

Note on the carried-forward Codex P2 thread: it's pinned to the original commit `8a334a45`, not the fix — the wizard-unmount race is resolved by the `mergeShots` snapshot (commit `0759f161`), which this PR's claude-review pass explicitly endorsed ("the snapshot pattern is the right call").

ted-design:
Round 3 — `256e72a5`. (Thanks for the architecture endorsement on the atomic batch + snapshot pattern.)

**Medium 1 — talent/talentIds lockstep.** Fixed. New `unionTalent()` keys the union by `talentId` so the same id yields one entry with the **primary's** name (first-seen wins), keeping `talent` and `talentIds` aligned; legacy name-only talent (no id) is carried through, name-deduped. +2 tests (divergent-name lockstep + legacy name-only).

**Medium 2 — `heroProductId`/`displayImageId` written as null.** Fixed: fall back to `undefined` (not `null`) when the id is no longer valid, so `sanitizeForFirestore` strips the key rather than writing a null to a never-set field. +1 test asserting the keys are **absent** on the invalid path.

**Low 4 — dangling notes divider.** Fixed: `notesAddendum` is only touched when the **secondary** contributes prose; when only the primary has an addendum it's left untouched (no trailing em-dash). Updated the old test + added a secondary-only case.

**Low 5 — wizard error-path test.** Added: `executeShotMerge` rejecting → `toast.error` fires, stays on the preview step, Merge button re-enabled.

**Low 3 — helper ordering.** `unionTalent` placed before its use; the remaining `union*` helpers are hoisted `function` declarations (work correctly) — left as-is to avoid churn.

**Low 6 — radiogroup arrow-key roving.** The ARIA roles are now correct (`role="radiogroup"`/`role="radio"`/`aria-checked`); full arrow-key roving for a 2-option internal picker is a reasonable follow-up rather than a blocker.

70 merge-suite tests green, build + lint clean.

ted-design:
Round 4 — `734c596f`. (Thanks for the approve.)

**Item 3 (must-fix) — custom radio primitive.** Fixed: the merge-mode selector now uses the Radix-backed shadcn `RadioGroup`/`RadioGroupItem` (`@/ui/radio-group`) — `sr-only` items + `<label>` cards preserve the card appearance and get native arrow-key roving (also closes round 3's keyboard-nav follow-up).

**Item 1 — "lockstep" wording.** Clarified the `unionTalent` comment: `talent`/`talentIds` are index-aligned only for the id-bearing prefix; legacy name-only entries (no `talentId`) are appended to `talent` beyond the `talentIds` length, so `talent.length > talentIds.length` for those shots. Documented the asymmetry explicitly.

**Items 4/5/6/8 — dispositioned:**
- Inline `bg-[var(--color-*)]`: the repo's established pattern across components (CastingCard, etc.); the named utilities aren't all defined in the Tailwind config, so the inline `var()` form is consistent here.
- `max-h-[calc(...)]`: no shared dialog-content max-height token exists; other dialogs use ad-hoc constraints similarly.
- `handleClose` also firing `onMerged`: intentional and correct (the 2-item selection is stale post-merge) — noted the naming.
- crew can merge: confirmed intended — `canManageShots` includes crew, matching the shots-write Firestore rules (CLAUDE.md "crew users have canManageShots").

**CI note:** a single `vitest` shard went red on `products/lib/shootReadiness.test.ts` (`sortByUrgency` tiebreak `['b','c']` vs `['c','b']`) — that's a **pre-existing, unrelated** flake: it passes 3/3 locally in isolation, this branch doesn't touch `products/`, and the parallel vitest shard + the `test` (E2E arbiter) + `claude-review` + `build` all passed on the same commit. Looks like a sort-stability / test-isolation issue in the products suite worth a separate follow-up.

70 merge-suite tests green, build + lint clean. I consider the review converged on this PR.

PR #449, opened 2026-06-12
feat(shots): quiet View-as (preview-as-crew) menu — Phase 5e-III

ted-design:
## Phase 5e-III — the quiet View-as menu

The final slice of the Shoot-surface redesign. A **global admin/producer** can **preview the Crew (Shoot) surface** without persisting anything, and return to their real view. Everything is gated behind `featureShootSurface` (the whole 5e initiative; **default OFF in prod** → this PR is a no-op until the flag flips).

### The hard invariants (pinned by tests)
1. **Never through `useEffectiveRole`.** That hook's module-level `resolvedRoleCache`, its commit, and the downgrade `toast(...)` would persist/announce a preview as a real role change. The file is **byte-untouched** (`git diff` empty). The preview interposes ONLY at surface resolution.
2. **Zero persistence** on enter/exit: no `:view:v1` write, no `setSearchParams`, no cache commit, no toast. The provider is in-memory `useState` only.
3. **Suppresses the url/stored rungs** (passes `urlView`/`urlGroup`/`storedView` as null) so the previewer's own stored `table` choice can't outrank — and mask — the previewed shoot-surface default.
4. **Narrows, never escalates.** Every real write-gate keeps using the real effective role; preview only swaps the presentation surface/affordances.

### What changed
- **`ViewAsPreviewProvider`** (new) — in-memory `useState` preview role, modeled on `SearchCommandProvider`; mounted in `App.tsx` so it survives list→detail nav.
- **`useResolvedSurface`** — at the existing `// 5e-III View-as seam`, `effectiveRole: previewRole ?? role` (url/stored already null). Drives the shell mount + `affordances`/`chrome`.
- **`useShotListState`** — `ShotListSurfaceContext.previewActive` (additive) nulls the view rungs in the pure `resolveSurface` derivation. `ShotListPage` surfaceContext sets `role: previewRole ?? role` (the page's real write-gates keep the real `role`).
- **`ViewAsMenu`** (new) — chip-adjacent dropdown in the shots `PageHeader`; self-gated on `featureShootSurface` + the GLOBAL admin/producer claim. "Crew (Shoot)" / "Return to your view".
- **`CommandPalette`** — "View as Crew (Shoot)" / "Return to your view" quick actions, same gate, no navigation.
- **`EffectiveRoleChip`** — distinct "Previewing as X" state (Eye + strong border), independent of the downgrade pill.

### Build approach
Scout → build → adversarial-verify dynamic workflow: 1 foundation writer + 4 parallel consumer writers + 3 adversarial verifiers (all PASS — `useEffectiveRole` untouched, narrow-not-escalate, gates+inertness). Gates re-verified on the main loop.

### Gates (verified on the main loop, independent of the workflow)
- **181 tests green** across the 12 touched/adjacent files (incl. the component-adjacent `FloatingActionBar.test.tsx`, the resolver inertness pins, and the Shoot shell tests).
- New coverage: provider zero-persistence spy; `useResolvedSurface` preview→`shoot` regardless of real role; chip preview state; `useShotListState.resolver` stored-`table` suppression under preview (`setSearchParams` count 0); palette quick actions.
- Lint clean · production build clean. `featureShootSurface` OFF → flag-off path byte-identical to 5e-II trunk.

### DIFF
--- a/src-vnext/app/App.tsx
+++ b/src-vnext/app/App.tsx
@@ -3,6 +3,8 @@ import { Toaster } from "sonner"
 import { AuthProvider } from "@/app/providers/AuthProvider"
 import { ThemeProvider } from "@/app/providers/ThemeProvider"
 import { SearchCommandProvider } from "@/app/providers/SearchCommandProvider"
+// 5e-III: in-memory "View as" preview provider (no persistence)
+import { ViewAsPreviewProvider } from "@/app/providers/ViewAsPreviewProvider"
 import { TooltipProvider } from "@/ui/tooltip"
 import { AppRoutes } from "@/app/routes"
 
@@ -12,10 +14,13 @@ export function App() {
       <ThemeProvider>
         <AuthProvider>
           <SearchCommandProvider>
-            <TooltipProvider delayDuration={200}>
-              <AppRoutes />
-              <Toaster position="bottom-right" richColors closeButton />
-            </TooltipProvider>
+            {/* 5e-III: View-as preview state — in-memory only, consumers under it need auth too */}
+            <ViewAsPreviewProvider>
+              <TooltipProvider delayDuration={200}>
+                <AppRoutes />
+                <Toaster position="bottom-right" richColors closeButton />
+              </TooltipProvider>
+            </ViewAsPreviewProvider>
           </SearchCommandProvider>
         </AuthProvider>
       </ThemeProvider>
--- a/src-vnext/app/App.tsx
+++ b/src-vnext/app/App.tsx
@@ -3,7 +3,6 @@ import { Toaster } from "sonner"
 import { AuthProvider } from "@/app/providers/AuthProvider"
 import { ThemeProvider } from "@/app/providers/ThemeProvider"
 import { SearchCommandProvider } from "@/app/providers/SearchCommandProvider"
-// 5e-III: in-memory "View as" preview provider (no persistence)
 import { ViewAsPreviewProvider } from "@/app/providers/ViewAsPreviewProvider"
 import { TooltipProvider } from "@/ui/tooltip"
 import { AppRoutes } from "@/app/routes"
@@ -14,7 +13,6 @@ export function App() {
       <ThemeProvider>
         <AuthProvider>
           <SearchCommandProvider>
-            {/* 5e-III: View-as preview state — in-memory only, consumers under it need auth too */}
             <ViewAsPreviewProvider>
               <TooltipProvider delayDuration={200}>
                 <AppRoutes />
--- a/src-vnext/app/providers/ViewAsPreviewProvider.tsx
+++ b/src-vnext/app/providers/ViewAsPreviewProvider.tsx
@@ -2,11 +2,13 @@ import {
   createContext,
   useCallback,
   useContext,
+  useEffect,
   useMemo,
   useState,
   type ReactNode,
 } from "react"
 import { Role } from "@/shared/types"
+import { useAuth } from "@/app/providers/AuthProvider"
 
 // 5e-III: In-memory "View as" preview. A GLOBAL admin/producer can preview the
 // Crew (Shoot) surface without persisting anything. NO localStorage, NO URL,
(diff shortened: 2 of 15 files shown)


ted-design:
## Review dispositions (claude-review + Codex)

Pushed two follow-up commits: `54600762` (dropdown close) and `25340e20` (Codex P2 + comment trims). All notes addressed below.

### Codex
- **P2 — preview survives auth identity/role change** → **FIXED** (`25340e20`). The provider now reads `useAuth` and resets `previewRole` whenever `user.uid` or the global `role` changes (in-place sign-out/sign-in, or a mid-session claim change). Reset only — still zero persistence (no localStorage/URL/cache/toast). +2 tests (user swap, claim change); provider suite 4→6 green. Good catch — the stuck-no-exit state was real.

### claude-review
1. **Dropdown doesn't close after selection** → **FIXED** (`54600762`). Dropped `event.preventDefault()` on both `DropdownMenuItem` `onSelect` handlers so Radix's default close fires (matching the palette's close-on-select). Caught independently in the :5174 eyeball. ViewAsMenu tests still 8/8.
2. **Comment discipline** → **PARTIALLY ADDRESSED**. CLAUDE.md's verifiable rule (line 100) is "don't comment code you didn't change" — these comments are all on changed lines, so they're compliant; I didn't find a "no multi-line comment blocks" rule in CLAUDE.md. That said, I trimmed the two most redundant blocks you named (the `<ViewAsMenu />` JSX comment and the `previewActive` JSDoc). The remaining comments document non-obvious security invariants (zero-persistence, never-through-`useEffectiveRole`, the url/stored-rung suppression) that a future edit could silently break — keeping those deliberately.
3. **Inline CSS-var classes in `EffectiveRoleChip`** → **DECLINED (false positive)**. The existing downgrade pill in the *same file* (line 42, unchanged) already uses `border-[var(--color-border)] bg-[var(--color-surface-subtle)] text-[var(--color-text-muted)]` — the preview chip matches that established neighbor pattern exactly (CLAUDE.md "match surrounding code"). Migrating to token-mapped classes would be a separate file-wide sweep, not a 5e-III concern.
4. **`useViewAsPreview()` outside provider fails silently** → **DECLINED (low value)**. The provider is mounted at app root (inside `AuthProvider`), so it's structurally always present; a dev-only warn would be dead code.
5. **Flag mock passthrough in `CommandPalette.test`** → **DECLINED (no behavioral impact)**. `CommandPalette` consumes only `featureShootSurface`; the mock is exhaustive for current behavior. An `importActual` passthrough is speculative for a flag the component never reads.

### Eyeball (producer→preview-as-crew on :5174, throwaway project, deleted after)
Menu renders (admin + flag); enter preview flips list→Shoot shell + chip "Previewing as Crew"; survives list→detail→back nav (ShootShotDetail w/ sticky tap-row); "Return to your view" restores the plan-build table; palette "View as Crew (Shoot)" works + closes. Q2-26 No. 3 never opened.

ted-design:
## Dispositions — claude-review final pass (`3b6d7bd7`)

Thanks — all five addressed; none was blocking (CI was green with them).

1. **`useMemo` missing `setPreviewRole` dep** → **FIXED**. Added it (always-stable setter; harmless).
2. **`EffectiveRoleChip` preview-branch ordering** → **FIXED**. Replaced the precedence comment with the actual WHY it sits ahead of the auth/role loading guards (a preview is only ever set by deliberate action, so on a mid-session claim refresh the previewer should still see the chip, not a blank).
3. **Over-commenting (App.tsx, ShotListPage)** → **FIXED**. Trimmed the three what-not-why `// 5e-III` comments. Kept the substantive invariant comments you flagged as justified (provider block, `useResolvedSurface` seam, `useShotListState` suppression rationale).
4. **`ViewAsMenu` disabled-item keyboard guard** → **FIXED**. `onSelect` now guards `if (isPreviewing) clearPreview()` + a test pinning **warehouse** is excluded from the gate (closing the coverage gap you noted).

Minor observations: `canViewAs` duplication left as-is per your own "three similar lines beats a premature abstraction" call. Provider-inside-`AuthProvider` nesting confirmed correct.

Gates re-verified locally: lint clean, ViewAsMenu 8→9, provider/chip/list suites green, build clean. Merging once CI is green.

PR #450, opened 2026-06-12
feat(rules): revoke warehouse lane-write — Phase 5f-I

ted-design:
## Phase 5f-I — warehouse lane-write revoke (backend only)

First PR of Phase 5f (client/warehouse read-only Review surfaces). **Backend only — no app/UI code.** Implements the locked **Q3** decision: warehouse moves to a read-only product-forward Review surface, so it can no longer write lane/scene data.

### The 3-edit revoke (the overlapping-OR trap)
Firestore ORs all matching rules, so narrowing the explicit `/lanes/{laneId}` arm alone is **cosmetic** — the wildcard catch-all re-grants it. All three edits are required:
1. lanes **create/update** arm (`firestore.rules:890`): `['producer','warehouse']` → `['producer']`
2. lanes **delete** arm (`:911`): `['producer','warehouse']` → `['producer']`
3. wildcard catch-all **write** arm: added `collectionId != 'lanes'` next to the existing `!= 'members'` guard. Surgical exclusion — the scout confirmed `crew` is the only sub-collection truly under the wildcard write, so zero collateral (vs. dropping `warehouse` from the list).

**Warehouse lane READ is preserved** (explicit `:887` + wildcard read arms) — load-bearing for the `review-warehouse` surface. The `direction`/`notes`/`sceneNumber` size-validation conjunction is byte-identical.

### Tests (emulator rules-lane)
`shots.rules.test.ts` gains `[w1]`–`[w5]` + a generalized wildcard-guard:
- `[w1]` warehouse can't create a lane (explicit arm) → DENY
- `[w2]` warehouse can't update a lane **via the wildcard path** → DENY — **the non-vacuous A3 regression-catcher** (it passes the write, and fails, if the wildcard exclusion is reverted)
- `[w3]` warehouse can still **read** a lane → ALLOW
- `[w4]` producer can still create/update/delete → ALLOW
- `[w5]` crew still can't write → DENY

Built via a scout→build→adversarial-verify workflow; verifier PASS on all six invariants; diff re-verified by hand. Lint clean; test file compiles (49 emulator-gated cases run in CI's `test` job).

### Sequencing note (read before merging/deploying)
- The `canEditScene` **UI** removal (spec edit A1) is **deliberately deferred to 5f-III** so the warehouse experience changes atomically at go-live.
[...]

PR #451, opened 2026-06-12
feat(shots): client Review surface (review-client) — Phase 5f-II

ted-design:
## Phase 5f-II — client Review surface (review-client)

Second PR of Phase 5f. **App code behind a new `featureReviewSurface` flag (default OFF)** — zero prod impact on merge. Builds the **client approval surface** the resolver already names (`viewer` → `review-client`) but which until now fell through to the full producer editor.

### What's built
- **`flags.ts`** — `featureReviewSurface` (default OFF, `VITE_REVIEW_SURFACE`), mirrors `featureShootSurface`. Mandatory because the resolver emits `review-client` unconditionally, so a fork keyed on the surface value alone would ship live without a kill switch.
- **`ReviewShotDetail.tsx`** (new) — one read-only shell, `variant` prop. `review-client`: large hero (`canUpload={false}`) → read-only status badge → **open comment composer** → read-only talent/location meta → `ProductColorwayStrip readOnly` last. **No version-history section, no status tap-row, no write handlers mounted.** (`review-warehouse` is a placeholder — 5f-III.)
- **`ReviewClientGallery.tsx`** (new) — image-led read-only tiles (hero + status + colorway chips + comment count), not the producer table/card fork. Matches the approved comp (`mockups/s26-5f-review-surfaces-comp.html`, Client surface).
- **Mount forks** in `ShotDetailPageUnified` + `ShotListPage` for `surface==='review-client'`, gated on the flag, beside the existing shoot forks. **Flag-off byte-identical.**
- **Q4 comment collapse (scoped)** — `ShotCommentsSection` gains an optional `writeAuthoritative` prop: when set, the composer is governed by `canComment` alone (the internal `canManageShots(role)` gate is bypassed). `ReviewShotDetail` passes it for clients; **both existing callers (producer editor, shoot shell) omit it → their double gate is byte-identical.** Rules already permit viewer comment-create — this only opens the UI, scoped to Review. Author-only update/delete unchanged.

### Tests / gates
`ReviewShotDetail` composition (hero-first, open composer, no version history), Q4 scoping (viewer enabled only with `writeAuthoritative`; producer/shoot unchanged), flag-OFF no-change, + emulator rules-lane `[c1]`–`[c4]` pinning the existing viewer comment rules. **107 tests pass + 47 emulator-gated; lint + production build clean.** Built scout→foundation→parallel-consumers→adversarial-verify; verifier PASS on all 7 invariants; diff re-verified by hand. `PublicShotSharePage` byte-untouched.

### Notes for review
- **`shots.rules.test.ts` overlap with #450 (5f-I)** — both touch it in separate regions; trivial reconcile for whichever merges second.
- **Live `:5174` eyeball pending** — deferred to a consolidated client+warehouse pass after 5f-III (the spec frames it as one role×surface matrix).
- `canEditScene`/warehouse changes are NOT here (5f-III). This PR is client-only.

Spec: `outputs/2026-06-12-phase5f-build-spec.md` §2/§5.

### DIFF
--- a/src-vnext/features/shots/__tests__/shots.rules.test.ts
+++ b/src-vnext/features/shots/__tests__/shots.rules.test.ts
@@ -101,6 +101,13 @@ const SHOT_ADMIN_PRIVATE_UPDATE = "shot-admin-private-update"
 const SHOT_ADMIN_EMPTY_UPDATE = "shot-admin-empty-update"
 const SHOT_VERSIONS = "shot-versions"
 const SHOT_VERSIONS_SEQ = "shot-versions-seq"
+// 5f-II Q4 — the shot whose comments thread the viewer-commenting cases use.
+const SHOT_REVIEW_COMMENTS = "shot-review-comments"
+// A comment authored by SOMEONE ELSE (the owner) — the viewer must not be able
+// to mutate/delete it ([c2]/[c3]). And the viewer's OWN comment — soft-delete
+// via the allowed update path is permitted ([c4]).
+const COMMENT_OTHERS = "comment-owned-by-other"
+const COMMENT_VIEWER_OWN = "comment-owned-by-viewer"
 
 /** Shot create payload — mirrors CreateShotDialog.tsx handleCreate. */
 function makeShotCreate(projectId: string, createdBy: string) {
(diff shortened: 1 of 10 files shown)


ted-design:
**Codex disposition (1 P2):**

- **"Hide producer actions on the review gallery"** — **valid, fixed** (also independently caught in the :5174 eyeball). The review forks only swapped the list *body*; the PageHeader Export/Share/Select/New Shot actions still rendered from capability flags. Fixed at the stacked tip **#452 `65fbc48c`** by guarding those header actions with `!isReviewSurface` (covers both review-client and review-warehouse in one place). Flag stays OFF in prod, so no user ever saw the leak.

PR #452, opened 2026-06-12
feat(shots): warehouse Review surface (review-warehouse) + canEditScene A1 — Phase 5f-III

ted-design:
## Phase 5f-III — warehouse Review surface (review-warehouse) + the Q3 UI revoke

**Stacked on #451 (5f-II)** — base is `feat/review-client-shell-phase5f-ii` so this diff is warehouse-only. App code behind the existing `featureReviewSurface` flag (default OFF). Builds the **product-forward warehouse pull/prep view** (Ted's pick off the real-data comp) + the **UI half of the Q3 scene-edit revoke**.

### What's built
- **`warehousePullList.ts`** (new, pure + 13 tests) — `buildWarehousePullList` aggregates **every** product requirement across **all looks of all shots** into a pull list keyed by family+colourway+size, each row carrying label / style·SKU / colour swatch / the shots that need it (deduped, stable-sorted). **No prep-status column** — no such field exists in the shot/product model (the only fulfilment state lives in the separate *pulls* feature), so the list aggregates honestly from real data rather than inventing the comp's mock prep column.
- **`ReviewWarehouseList.tsx`** (new) — read-only product-led pull list (swatch + label + size + style/SKU + "needed for" shot chips). No bulk/quick-add/FAB.
- **`ReviewShotDetail.tsx`** — `review-warehouse` variant: `ProductColorwayStrip readOnly` **first** (product-forward) → read-only status header → read-only meta → **comment composer ON** (`writeAuthoritative`, warehouse flags product issues). No version history, no scene/lane edit, no status tap-row.
- **Forks** in `ShotListPage` + `ShotDetailPageUnified` extended from review-client to also handle `review-warehouse`, flag-gated. **Flag-off byte-identical.**
- **`rbac.ts` `canEditScene` (A1)** — remove `ROLE.WAREHOUSE` → admin/producer only. **Scoped to scene/lane-edit UI** (`SceneDetailSheet`); **warehouse KEEPS pulls** (`canManagePulls` untouched, pinned by a test).

### The Q3 split (read before merge/deploy)
A1 here is the **UI half** of the warehouse revoke; the **backend half** (lanes rule + wildcard catch-all) is **5f-I / #450**. They align on `main` once both merge. Deploy the rules (`firebase deploy --only firestore:rules`) on your go, **coordinated with the `featureReviewSurface` flip**, so warehouse never sees a UI/rules-mismatch window. (An adversarial verifier on this stacked branch correctly noted the lanes rule still grants warehouse here — that's because #450 is the parallel branch; it's the intended split, not a gap.)

### Gates
Lint clean; **138 tests** (pull-list aggregation incl. multi-shot dedup + non-primary looks, warehouse composition, A1 + warehouse-keeps-pulls, flag-off no-change) + fork regressions; production build clean. Built foundation→parallel-consumers→adversarial-verify; verifier PASS on all 10 invariants. `PublicShotSharePage` byte-untouched.

### Notes
- **Live `:5174` eyeball (client+warehouse) pending** — the remaining pre-merge gate.
- `allowExports` per-share toggle is **NOT here** — a tiny final 5f-IV.
- Merge order: #450 (5f-I rules) and #451 (5f-II client) first, then this.

Spec: `outputs/2026-06-12-phase5f-build-spec.md` §2.

### DIFF
--- a/src-vnext/features/shots/components/ReviewShotDetail.tsx
+++ b/src-vnext/features/shots/components/ReviewShotDetail.tsx
@@ -19,10 +19,18 @@
 //   → minimal read-only meta (talent / location) → products read-only LAST
 //   (ProductColorwayStrip readOnly=true).
 //
-// Deliberately ABSENT (scoping boundaries): no version-history section
-// (edit-diff audit is a producer/crew tool, not a client approval affordance),
-// no field editors, no lifecycle menu, no status tap-row (clients can't write
-// shots — status is shown read-only), no upload, no share.
+// review-warehouse composition (the warehouse's job is to PULL): product-FORWARD
+//   — ProductColorwayStrip readOnly FIRST (every look's garments, because the
+//   warehouse pulls everything, not just the cover) → shot # / title /
+//   read-only status context → read-only talent / location meta → comment
+//   composer OPEN (same writeAuthoritative escape hatch as the client variant;
+//   warehouse can leave notes). NO hero-led layout, NO prep status (no such
+//   field exists in the model).
+//
+// Deliberately ABSENT (scoping boundaries, BOTH variants): no version-history
+// section (edit-diff audit is a producer/crew tool, not a review affordance),
+// no field editors, no lifecycle menu, no status tap-row (review surfaces don't
+// write shots — status is shown read-only), no upload, no share.
 import { useMemo } from "react"
 import { useParams } from "react-router-dom"
 import { LoadingState } from "@/shared/components/LoadingState"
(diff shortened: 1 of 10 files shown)


ted-design:
**Codex disposition (2 P2 — both fixed in `65fbc48c`):**

1. **"Use unfiltered shots for warehouse pulls"** — fixed. Both review lists now feed from `unfilteredSortedShots` (not `displayShots`). The filter toolbar is suppressed on these surfaces, so a lingering deep-link filter would otherwise silently drop required items with no control to clear it — the exact Shoot-shell hazard, fixed the same way.
2. **"Include legacy shot.products in pull list"** — fixed. `buildWarehousePullList` now falls back to `extractShotAssignedProducts` when a shot has no look products, so legacy/imported shots' top-level assignments aren't silently omitted; dedup-by-compositeKey keeps it safe. +2 unit tests (legacy-only shot → row; looks+legacy → no double-count).

Plus the eyeball-driven header-suppression fix (Codex #451) landed here too. Lint + 69 targeted tests + build green.

PR #460, opened 2026-06-16
feat(talent): detail IA re-group — contact meta-line + Fit-signals/Creative-assets zones — Talent redesign Phase 2

ted-design:
## Talent redesign — Phase 2 (detail IA)

Re-groups the talent **detail panel's Profile tab** from a flat 6-section stack into a tiered information architecture, behind a new feature flag. Continues the Talent redesign after 0a (#456), 0b (#457), 1a (#458), 1b-i (#459).

**Flag:** `featureTalentDetailIA` (env `VITE_TALENT_DETAIL_IA`, **default OFF**). The flag-off render path is **byte-identical to trunk** — verified by a normalized zero-token diff of the `else` branch against `main`.

### What flag-ON does (per the locked spec decisions, `outputs/2026-06-13-talent-redesign-build-spec.md` §7 + §5)
- **Contact → inline editable meta-line** (locked Decision #3). New `TalentContactMetaLine` renders `Email · Phone · Web` as one editorial line below the hero. Each field stays individually editable via the same `InlineEdit` + the now-exported `ReferenceUrlField` (no duplication); `savePatch` wiring is byte-identical to the old card.
- **Measurements + Notes → "Fit signals" zone** (locked Decision #4) — the "does she fit?" group, ahead of creative assets.
- **Portfolio + Casting → "Creative assets" zone** — the image-heavy group (§5 goal).
- **Projects → read-only tag row** in a meta/actions zone (demoted from a bordered card; links preserved; hidden when empty).
- **Delete** preserved, isolated in the meta/actions zone.
- **Hero (gender badge), the Shot History tab, and all sub-component internals are untouched.**

Zones use `heading-section` (16px) eyebrows over the cards' existing `heading-subsection` (14px) — correct two-tier hierarchy, token-backed.

### Verification
- **Targeted tests** ✅ — flags 13, `LibraryTalentPage.detailIA.test.tsx` 3 (new), `rosterIA` 3 (unbroken), main `LibraryTalentPage` 8 (incl. a new flag-OFF byte-identical assertion).
- **Full suite** ✅ — 275 files / 3707 tests pass; 79 skips = the documented quarantine (unchanged).
- **tsc** ✅ — 160 = baseline exactly, **0 in changed files** (delta 0).
- **3-lens adversarial verify** ✅ PASS on all three (flag-off byte-identical · faithful-IA/decisions · hand-audit/regression). No correctness defects.

### Flag-ON eyeball items for Ted (at flip time, not blockers)
[...]

PR #461, opened 2026-06-16
test(products): de-flake sortByUrgency (shared now → identical same-window launch dates)

ted-design:
## De-flake `sortByUrgency`

The `shootReadiness.test.ts > sortByUrgency` case intermittently failed in CI (e.g. on Talent PR #460's `vitest` job) — `expected ['b','c','a'] to deeply equal ['c','b','a']`.

**Root cause (test-only, not the impl):** the two items intended to share a launch window (`b`, `c`) each call the `Date.now()`-relative `ts(5 * DAY_MS)` helper. When the millisecond clock ticks between constructing `b` and `c`, their launch dates differ by ~1ms, so the date comparison (not the readiness tie-break) decides their order → `b` (earlier date) sorts before `c` → `['b','c','a']`. When both land in the same millisecond (the common case), the dates tie and the readiness tie-break correctly yields `['c','b','a']`.

**Fix:** capture `now` once and build all three dates from it, so `b` and `c` get a byte-identical `launchDate` and the readiness tie-break applies deterministically. The `sortByUrgency` implementation is **unchanged** — it was correct; only the test's date construction was non-deterministic.

Verified deterministic across 5 consecutive runs (20/20 each).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src-vnext/features/products/lib/shootReadiness.test.ts
+++ b/src-vnext/features/products/lib/shootReadiness.test.ts
@@ -224,10 +224,13 @@ describe("computeSuggestedShootWindow", () => {
 
 describe("sortByUrgency", () => {
   it("sorts by launch date ascending, then readiness ascending", () => {
+    // Shared `now` so the two same-window items get an identical launch date.
+    const now = Date.now()
+    const at = (offsetMs: number) => Timestamp.fromDate(new Date(now + offsetMs))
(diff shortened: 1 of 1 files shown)


ted-design:
@codex review

PR #462, opened 2026-06-16
feat(talent): suppress empty contact fields for read-only viewers in the detail-IA meta-line

ted-design:
## Phase 2 detail-IA — polish follow-up

Small follow-up to PR #460, behind the same `featureTalentDetailIA` flag (default OFF). Addresses the empty-field item flagged at #460 review/eyeball time. (Ted confirmed the URL label stays **"Web"**.)

**Empty-field suppression (read-only viewers).** The meta-line previously always rendered all three segments + their separator dots, so a viewer of a talent with no contact info saw `Email — · Phone — · Web —`. Now:
- **Editors** still see all three fields (so blanks stay clickable to fill in).
- **Read-only viewers** only see fields that actually have a value, with a `No contact details` fallback when there are none — and no orphan separator dots.

### Scope / safety
- **Flag-on only.** `TalentContactMetaLine` renders only in the flag-on detail path; the flag-off `else` branch in `TalentDetailPanel` is **untouched** (still byte-identical to trunk).
- New `TalentContactMetaLine.test.tsx` (4 cases): editor-sees-all-when-empty · viewer-suppresses-empties · read-only-all-empty placeholder · exactly-one-separator-between-two-shown-fields.
- Full suite green; tsc delta 0.

### Deferred (not in this PR)
The edit-mode input width/reflow is the one inherently-visual item. An initial `max-w-[220px]` was removed because `DESIGN_SYSTEM.md:219` forbids hardcoded `max-w-[Npx]` on text fields, and the value couldn't be verified without rendering the flag-on path. It's deferred to the flag-on `:5174` eyeball, where a layout-based fix can be confirmed.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src-vnext/features/library/components/TalentContactMetaLine.tsx
+++ b/src-vnext/features/library/components/TalentContactMetaLine.tsx
@@ -11,12 +11,11 @@ interface TalentContactMetaLineProps {
 }
 
 // One inline editorial meta-line segment: an uppercase micro-label + its value.
-// The value is width-capped so the edit-mode input can't balloon the wrapping row.
 function Segment({ label, children }: { readonly label: string; readonly children: ReactNode }) {
   return (
     <span className="inline-flex min-w-0 items-baseline gap-1.5">
       <span className="label-meta shrink-0">{label}</span>
-      <span className="min-w-0 max-w-[220px]">{children}</span>
+      <span className="min-w-0">{children}</span>
     </span>
   )
 }
--- a/src-vnext/features/library/components/TalentContactMetaLine.tsx
+++ b/src-vnext/features/library/components/TalentContactMetaLine.tsx
@@ -1,4 +1,4 @@
-import type { ReactNode } from "react"
+import { Fragment, type ReactNode } from "react"
 import { InlineEdit } from "@/shared/components/InlineEdit"
 import { ReferenceUrlField } from "@/features/library/components/TalentContactSection"
 import type { TalentRecord } from "@/shared/types"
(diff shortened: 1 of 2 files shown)


ted-design:
@codex review

ted-design:
Good catch — you're right, `DESIGN_SYSTEM.md:219` (§ Table Patterns) explicitly forbids hardcoded `max-w-[Npx]` on text fields. Removed it in the follow-up commit.

Rather than swap in an unverified layout cap, I've **scoped this PR to the empty-field suppression only** (the tested, rule-clean half). The edit-mode reflow is the one inherently-visual item — I can't drive the Firebase-authed app to see whether the input actually balloons or how a `min-w-0`/`flex-1` approach reads — so it's deferred to the flag-on `:5174` eyeball, where it can be seen and fixed with a layout-based constraint with visual confirmation.

Dispositioned the two non-blocking notes (the `--color-text-subtle` token usage and the "manual array entry if a 4th field is added") — agreed, no change needed.

ted-design:
Acted on the two-field test suggestion (`07…` follow-up) — it asserts exactly one separator between two shown viewer fields, locking the interleaving between the existing zero-dot and two-dot cases.

Declining the semantic-Tailwind-class suggestion for the `No contact details` span: `text-[var(--color-text-subtle)]` is the DESIGN_SYSTEM-mandated CSS-token pattern and (per your own earlier review on this PR) `text-subtle` has no Tailwind alias — so there's no semantic class to swap to. No change.

PR #463, opened 2026-06-16
fix(shots): show explicit cover product's catalog image when its URL wasn't denormalized

ted-design:
## Bug (confirmed regression)
A product **with a thumbnail**, selected as a look's **cover product** in the shots editor, renders **blank** as the cover — while the product still shows its thumbnail in the product row. Reported as "it used to show."

## Root cause
`mapShot.normalizeHeroImage` is a pure sync mapper that resolves a cover product's image **only** from denormalized URL fields on the look's assignment (`skuImageUrl/thumbUrl/familyImageUrl`); it has **no catalog fallback**. Commits `73af543f` ("make cover product selection explicit") and `bd9a1160` ("improve cover image rendering and clearing") — both intentional UX fixes — removed the older fallback chain so an explicit cover shows *exactly* the chosen product and a cleared cover stays cleared. The unhandled edge: a cover product whose catalog image was **never denormalized onto the assignment** (legacy assignment, or a catalog image added *after* the product was assigned — editing doesn't backfill it) resolves to `undefined`, so the cover is blank. The product **row** still shows the thumbnail because `AssignmentRow` (`1900bf28`) *does* fall back to the catalog family/sku doc.

> Note: the audit's first hypothesis (editing a cover product wipes its image URLs) was **refuted** — `handleConfirm`'s patch omits undefined image keys (`if (thumbUrl) …`), so the edit spread *preserves* the existing URL. The real fault is the render path's missing fallback.

## Fix
- New pure helper `coverProductImage.ts` `findExplicitCoverAssignment(shot)` — returns the assignment explicitly chosen as the cover, faithfully mirroring `normalizeHeroImage`'s resolution order (active-look **lock**; stale activeLookId falls through; first explicit look otherwise; match by sku/colour/family id).
- `HeroImageSection` — when `heroImage` is unresolved **and** a look has an explicit cover product, resolves that product's catalog image (mirroring `AssignmentRow`'s exact `thumbUrl ?? skuImageUrl ?? familyImageUrl ?? sku.imagePath ?? family.thumbnailImagePath ?? family.headerImagePath` chain via `useProductFamilyDoc`/`useProductSkuDoc` + `useStorageUrl`).
- **Does not revert** the Feb-4 intent: cover-cleared → still blank; explicit cover → only the chosen product; a resolved `heroImage` always wins.

## Tests
- `coverProductImage.test.ts` (10) — active-look lock, cleared/auto, match by family/sku/colour, stale-active fall-through, first-explicit-wins, no-match.
- `HeroImageSection.test.tsx` (+4) — renders the cover product's catalog image on fallback; no fallback when cleared; graceful empty-state when no catalog image; resolved hero image wins. (The new fallback test fails against pre-fix code — load-bearing.)
- Full suite: **277 files / 3725 tests pass** (79 pre-existing QUARANTINE skips), tsc delta 0.

## Scope
Editor hero only (the reported bug). The shot **grid card** (`ShotCard.tsx`) and **table thumbnail column** (`shotColumnRenderers.tsx`) share the same blank-cover root cause and are a **tracked follow-up** (shared cover-thumb component; the table cell needs a render-fn→component wrap).

### DIFF
--- a/src-vnext/features/shots/components/HeroImageSection.tsx
+++ b/src-vnext/features/shots/components/HeroImageSection.tsx
@@ -4,6 +4,8 @@ import { updateShotWithVersion } from "@/features/shots/lib/updateShotWithVersio
 import { useAuth } from "@/app/providers/AuthProvider"
 import { Button } from "@/ui/button"
 import { useStorageUrl } from "@/shared/hooks/useStorageUrl"
+import { useProductFamilyDoc, useProductSkuDoc } from "@/features/shots/hooks/usePickerData"
+import { findExplicitCoverAssignment } from "@/features/shots/lib/coverProductImage"
 import { ImagePlus, Loader2, RotateCcw } from "lucide-react"
 import { toast } from "sonner"
 import type { Shot } from "@/shared/types"
@@ -41,7 +43,22 @@ export function HeroImageSection({
   const fileRef = useRef<HTMLInputElement>(null)
   const [uploading, setUploading] = useState(false)
 
-  const heroCandidate = heroImage?.downloadURL ?? heroImage?.path
+  // Catalog-image fallback for an explicit cover product whose URL was never denormalized.
+  const coverAssignment = heroImage ? null : findExplicitCoverAssignment(shot)
+  const coverFamilyId = coverAssignment?.familyId ?? null
+  const coverSkuId = coverAssignment ? (coverAssignment.skuId ?? coverAssignment.colourId ?? null) : null
+  const { data: coverFamily } = useProductFamilyDoc(coverFamilyId)
+  const { data: coverSku } = useProductSkuDoc(coverFamilyId, coverSkuId)
+  const coverFallbackSrc = coverAssignment
+    ? (coverAssignment.thumbUrl ??
+       coverAssignment.skuImageUrl ??
+       coverAssignment.familyImageUrl ??
+       coverSku?.imagePath ??
+       coverFamily?.thumbnailImagePath ??
+       coverFamily?.headerImagePath)
+    : undefined
+
+  const heroCandidate = heroImage?.downloadURL ?? heroImage?.path ?? coverFallbackSrc
   const resolvedHeroUrl = useStorageUrl(heroCandidate)
   const canResetManual = !!heroImage?.path && heroImage.path.includes("/hero.webp")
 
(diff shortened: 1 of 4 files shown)


ted-design:
Thanks both — addressed in `fff0e16c`:

- **Codex P2 (stale catalog-doc flash) + claude-review #1 (unnecessary subscriptions):** fixed together. The family/sku docs are now only read when the assignment has **no** denormalized URL (the actual bug case), and each doc is **id-guarded** (`coverFamily?.id === coverFamilyId`) so a prior product's image can't feed the fallback while a new read is in flight or if it errors. Added a `ignores stale catalog doc data` test.
- **claude-review #2 (`colourImagePath`):** dispositioned. `ProductAssignment` has no typed `colourImagePath` (only `thumbUrl`/`skuImageUrl`/`familyImageUrl`) — `colourImagePath` exists only on raw Firestore data read by `normalizeHeroImage`. This chain intentionally mirrors `AssignmentRow`'s typed chain, and the catalog lookup (`sku.imagePath` = the colourway image) covers the legacy-image case.
- **claude-review #3 (document the auto-mode divergence):** dispositioned. The helper is named/documented as the *explicit* cover only, and the active-look lock + auto-mode behavior is covered by the unit tests. The suggested comment references the mapper cross-function, which the repo comment rule discourages.

ted-design:
@codex review

ted-design:
Re the latest review — all three are false positives against the **normalized** (typed) data the helper actually receives. `findExplicitCoverAssignment` operates on the mapped `Shot.looks[].products` (`ProductAssignment[]`), not raw Firestore docs, and `normalizeProducts` (`mapShot.ts:38-61`) builds a fresh typed object that has already folded the legacy fields:

- **Finding 1 — `productId`:** `mapShot.ts:47` sets `familyId: ((p.familyId ?? p.productId) as string) ?? ""`. Legacy `productId` is folded into `familyId` at map time, and the constructed object carries **no** `productId` (it's not a field on `ProductAssignment`, and the mapper doesn't spread raw `p`). So `p.familyId === heroId` already **is** `(p.familyId ?? p.productId) === heroId` — the suggested `?? p.productId` would be a type error and is redundant. No legacy shot is missed.
- **Finding 2 — `colourImagePath`/`thumbnailImagePath`:** `mapShot.ts:57-59` fold both legacy paths into `thumbUrl`/`skuImageUrl`/`familyImageUrl`. The 3-field `coverDenormalized` chain therefore already captures them; an assignment with a legacy path but none of the three typed URLs cannot exist post-normalization, so the "spurious read pair" can't occur.
- **Finding 3 — hero-wins test:** when `heroImage` is present, `coverAssignment = heroImage ? null : …` is `null`, so `coverFamilyId` is `null` and the fallback is short-circuited at the top — not by the id-guard. The test correctly asserts the resolved hero image is used; the two paths are mutually exclusive by construction, so there's no runtime ordering to assert.

No code change needed — the fix is correct and complete for normalized data. Happy to add a clarifying one-liner if preferred, though the repo comment rule discourages cross-function references.

ted-design:
@codex review

PR #464, opened 2026-06-16
feat(shots): scope talent filter to project talent + search, behind featureShotFilterTalentScope

ted-design:
## Problem
The shot-list advanced **"filter by talent"** picker (`FilterValuePicker.tsx` `TalentPicker`) listed the **entire org talent library** with **no search box** — you just scrolled all of them.

## Fix (behind `featureShotFilterTalentScope`, default OFF)
- **Scopes** options to **project-attached talent** (`talent.projectIds.includes(projectId)`) — the only project-attachment mechanism in the codebase, identical to what the shot-edit `TalentPicker` already shows.
- Adds a **search box**.
- **Two groups:** "Project talent" (always shown) + "All talent" (revealed only when searching, so the full library stays reachable without scrolling it by default).
- Empty states: "No talent assigned to this project" / "No project talent matches" / "No other talent matches".
- `projectId` threaded `ShotListPage → ShotListToolbar → ShotListFilterContent → FilterConditionRow → FilterValuePicker`; `talentRecords` type widened to carry `projectIds` (already present at runtime via `useTalent()`).

## Flag-off = byte-identical
With the flag OFF (prod default — deploy env doesn't define `VITE_SHOT_FILTER_TALENT_SCOPE`), `TalentPicker` early-returns the exact prior `CheckboxList` (full library, no search). The threaded `projectId` / widened type are inert. **The flag-on UX is an eyeball-then-flip item** (Firebase-authed; can't be driven solo).

## Decisions (Ted, locked)
A1 = `projectIds` scope · A2 = two groups (Project / All) · A3 = empty-state + searchable All.

## Tests
- `FilterValuePicker.test.tsx` (new, 5): flag-off byte-identical (full list, no search); flag-on scope-to-project; reveal-on-search; empty hint; toggle. All load-bearing (fail if the new code is reverted).
- `flags.test.ts` (+3): default off, `1`/`true` enables, other values stay off.
- Full suite **278 files / 3734 pass** (79 pre-existing skips), tsc delta 0.

### DIFF
--- a/src-vnext/features/shots/components/FilterConditionRow.tsx
+++ b/src-vnext/features/shots/components/FilterConditionRow.tsx
@@ -26,9 +26,10 @@ interface FilterConditionRowProps {
   readonly onRemove: (conditionId: string) => void
   readonly statusOptions: readonly { value: string; label: string }[]
   readonly tagOptions: readonly { id: string; label: string }[]
-  readonly talentRecords: readonly { id: string; name: string }[]
+  readonly talentRecords: readonly { id: string; name: string; projectIds?: readonly string[] }[]
   readonly locationRecords: readonly { id: string; name: string }[]
   readonly productFamilies: readonly { id: string; styleName: string }[]
+  readonly projectId: string
 }
 
 // ---------------------------------------------------------------------------
@@ -44,6 +45,7 @@ export function FilterConditionRow({
   talentRecords,
   locationRecords,
   productFamilies,
+  projectId,
 }: FilterConditionRowProps) {
   const meta = FILTER_FIELD_BY_KEY.get(condition.field)
   if (!meta) return null
@@ -118,6 +120,7 @@ export function FilterConditionRow({
         talentRecords={talentRecords}
         locationRecords={locationRecords}
         productFamilies={productFamilies}
+        projectId={projectId}
       />
     </div>
   )
--- a/src-vnext/features/shots/components/FilterValuePicker.tsx
+++ b/src-vnext/features/shots/components/FilterValuePicker.tsx
@@ -200,7 +200,7 @@ function TalentPicker({
         />
         {showOtherGroup ? (
           <TalentGroup
-            label="All talent"
+            label="Other talent"
             options={otherTalent}
             selected={selected}
             onToggle={onToggle}
--- a/src-vnext/features/shots/components/FilterValuePicker.tsx
+++ b/src-vnext/features/shots/components/FilterValuePicker.tsx
@@ -172,11 +172,14 @@ function TalentPicker({
   const onToggle = (id: string) => onChange(toggleInArray(selected, id))
 
   const q = query.trim().toLowerCase()
-  const matches = (name: string) => q === "" || name.toLowerCase().includes(q)
   const isProjectTalent = (t: { projectIds?: readonly string[] }) => t.projectIds?.includes(projectId) ?? false
+  const nameMatches = (t: { name: string }) => t.name.toLowerCase().includes(q)
+  const isSelected = (t: { id: string }) => selected.includes(t.id)
 
-  const projectTalent = talentRecords.filter((t) => isProjectTalent(t) && matches(t.name))
-  const otherTalent = talentRecords.filter((t) => !isProjectTalent(t) && matches(t.name))
+  // Selected talent stay visible regardless of search/scope — you can't uncheck what you can't see.
+  const projectTalent = talentRecords.filter((t) => isProjectTalent(t) && (q === "" || nameMatches(t) || isSelected(t)))
+  const otherTalent = talentRecords.filter((t) => !isProjectTalent(t) && ((q !== "" && nameMatches(t)) || isSelected(t)))
+  const showOtherGroup = q !== "" || otherTalent.length > 0
 
   return (
     <div className="space-y-2">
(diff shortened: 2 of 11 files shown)


ted-design:
Addressed in `526c6dc3`:

- **Codex P2 (search text resets after each toggle):** real bug — confirmed root cause. `deserializeFilters` assigns a fresh `crypto.randomUUID()` per condition (`filterSerializer.ts:74/105`) and `conditions` is re-derived from the URL on every update (`useShotListState.ts:169-172`), so the `condition.id`-keyed row remounted on each toggle and the local `query` reset. **Fixed by keying `FilterConditionRow` by `condition.field`** (stable; `AddFilterMenu` enforces one condition per field, so it's unique) instead of the ephemeral id — no remount on value change, query persists. General improvement for all pickers.
- **claude-review #1 (selected-but-hidden talent on flag flip):** fixed — `TalentPicker` now keeps **selected** talent visible regardless of scope/search (so an off-project selection always shows and can be unchecked). New test `keeps a selected off-project talent visible with an empty query`.
- **claude-review #2 (toggle test click target):** fixed — uses `getByRole("checkbox", { name })`.
- **claude-review #3 (missing deselect test):** added `deselects an already-selected talent`.
- **claude-review #4 (CSS-var token in `[]`):** dispositioned — `text-[var(--color-*)]` is the established codebase convention (design-tokens.js exposes CSS vars; no semantic class for `text-subtle`), and as noted it's not introduced by this PR. The minor observations (max-h-44, inline `isFeatureEnabled`, `isProjectTalent` nullish handling) are all correct as-is.

Full suite 278✓/3736✓, tsc delta 0.

ted-design:
Resolved the contradiction between the two reviews at the root, in `78f4bee2`:

The disagreement — claude-review wanted `key={condition.id}` (back), Codex's P2 was *caused by* `condition.id`. Both were right about a symptom: `condition.id` is the correct thing to key by **but** it wasn't stable — `deserializeFilters` minted a fresh `crypto.randomUUID()` per condition on every URL change (`filterSerializer.ts`), so the id-keyed row remounted on each toggle (Codex P2), while `condition.field` keying relied on an external uniqueness invariant (claude-review's fair concern).

**Fix:** `deserializeFilters` now mints a **deterministic** id (`\`${field}.${operator}\``) — unique per condition (the URL can't carry two of the same field+operator; `AddFilterMenu` also enforces one condition per field) and **stable across re-deserialization**. So:
- `key={condition.id}` is restored (claude-review) — and is now genuinely stable + unique.
- The row no longer remounts on value updates, so the talent search query persists (Codex P2 fixed at the source, not worked around).

3 new `filterSerializer` tests lock the stable-id invariant (identical ids across calls; id = field.operator; uniqueness within a result set). Full suite 278✓/3739✓, tsc delta 0. The PR description's "5 tests" is now 7 (added deselect + selected-visible).

ted-design:
@codex review

ted-design:
Both nits addressed in `3eaacc96`: (#2) `deserializeFilters` now skips duplicate `field.operator` segments via a `seen` set — a crafted URL can't cause an id/key collision; (#1) `addCondition` mints the same `${field}.${operator}` id, so every condition shares one scheme. New dedup test. Full suite 278✓/3740✓, tsc delta 0.

ted-design:
Re the ID-collision flag — it's not reachable in a rendered state (the rendered `conditions` always come from `useShotListState`'s `useMemo(deserializeFilters(searchParams))`, which now dedups; the `addCondition` array is only ever serialized, never rendered; and no caller adds a duplicate field — `AddFilterMenu` excludes active fields). The deterministic-id alignment of `addCondition` only *introduced* that theoretical surface for no functional gain, so I reverted it to the original transient UUID in `ec454014`. Net final state:

- `deserializeFilters` → **stable, deterministic, deduped** ids (the rendered conditions) — fixes the remount/query-reset (Codex P2) at the root and can't collide.
- `addCondition` → transient UUID (dropped on serialize; never rendered).
- `key={condition.id}` (stable + unique).

Full suite 278✓/3740✓, tsc delta 0.

ted-design:
Addressed both minor items in the latest push: group label `All talent` → `Other talent` (precise — it's the non-project group; matches the hint wording), and removed the stray comment so `addCondition` is byte-identical to trunk. @codex review

ted-design:
Good catch, Codex — you were right. `AddFilterMenu` only blocks `SINGLE_USE_FIELDS`; **Talent/Tag/Location/Product are multi-use** (repeatable), so my deterministic `field.operator` id collided on them and the dedupe dropped legitimately-repeated filters. I've **reverted the entire filter condition-id experiment** (`92bc9981`) — `deserializeFilters` is back to the original per-call random UUID with no dedupe (byte-identical to trunk), and `useShotListState`'s `addCondition` is unchanged. Added a `filterSerializer` test that **locks** the repeatable-field behavior (two `talent.in` rows → two distinct conditions with unique ids) to prevent a regression here.

Net PR is now the focused change: the scoped + searchable + grouped talent picker (`FilterValuePicker`), behind `featureShotFilterTalentScope` (default OFF, flag-off byte-identical), with `projectId` threaded through. `filterSerializer.ts` and `useShotListState.ts` carry **no** net change.

**Known follow-up (flag-ON only, documented):** the talent search query resets when you toggle a talent, because filter rows are keyed by `condition.id` and ids regenerate on each URL round-trip (pre-existing architecture). It's flag-OFF in prod; fixing it correctly needs a multi-use-safe stable id (e.g. positional) or lifting the query — deferred to the flag-on work rather than risk the multi-use bug again.

Re claude-review's `toggleInArray` flag: it's a **local function** (`FilterValuePicker.tsx:45`), not an import — no import to add. Full suite 278✓/3738✓, tsc delta 0.

ted-design:
@codex review

PR #465, opened 2026-06-16
feat(talent): controlled agency input (flag-gated) + normalization script

ted-design:
Maintenance item **B** (the last of Session 28's 3 flagged items). Closes the talent agency free-text → duplicate-variants gap with a controlled input + a one-time normalization migration. Built scout → build → adversarial-verify.

## What changed
**1. `AgencyCombobox` (new, cmdk `Command` + `Popover`)** — single-value, type-to-filter agency picker. Shows existing agencies first (to curb dupes), **allows free entry** (B2), **trims on write** (`updateTalent` doesn't trim), and **no-write-on-unchanged** (mirrors the `InlineEdit` guard it replaces). Replaces the free-text agency `InlineEdit` on `TalentHeroZone` **and** the agency `Input` on `CreateTalentDialog`.

**2. Threading** — `knownAgencies` (= `extractUniqueAgencies(talent)`, **memoized** so `React.memo(TalentDetailPanel)` stays stable) from `LibraryTalentPage` → `TalentDetailPanel` → `TalentHeroZone`, and into `CreateTalentDialog`.

**3. `scripts/normalize-agency-names.ts` (new)** — modeled on the proven `scripts/fixup-talent-gender.ts` (Firebase Admin, `--clientId`, dry-run default, chunked + 50ms). Sequencing:
- **Dump** (dry-run, no mapping): prints every distinct agency + count → Ted curates a `{variant → canonical}` JSON.
- **Preview** (`--mapping=…`): shows planned renames.
- **Write** (`--write`, requires `--mapping`): applies the reviewed map. Unmapped agencies left untouched + reported; map a variant to `""` to clear.

## Flag-gating
Behind **`featureTalentAgencyCombobox`** (env `VITE_TALENT_AGENCY_COMBOBOX`, **default OFF**; deploy-live doesn't define it). **Flag-off path is byte-identical** to trunk on both surfaces (the `InlineEdit` / `Input`).

## 🚩 Sequencing (Ted owns these op steps)
The agency filter `matchesAgency` (`talentFilters.ts`) is **exact-equality**, so variants only collapse in the filter once stored data is normalized:
1. Run the dump (`--clientId=…`) → curate the canonical map.
2. Preview (`--mapping`) → `--write`.
3. **Then** flip the flag (flag-on `:5174` eyeball is yours).

## Adversarial verification (4 lenses + confirm pass)
- **Flag-off byte-identical:** no findings.
- **Correctness (low, fixed):** re-selecting the current agency fired a redundant Firestore write the `InlineEdit` suppressed → added a no-write-on-unchanged guard (+ test).
- **Migration (low, fixed):** `loadMapping` silently dropped a row on a normalized-key collision with differing canonicals → now hard-stops at load time (before any Firestore connection).
- **Migration (nit, fixed):** canonical write value now whitespace-normalized (not just trimmed) so a sloppy canonical can't reintroduce a near-dup.
- **Constraints (false-positive, dismissed):** "empty-state never shows" — proven unreachable branch.

## Gates (verified locally)
- **Tests:** full suite **3755 passed, 79 pre-existing QUARANTINE skips, 0 fail**; **12 new** (10 `AgencyCombobox` + 2 `TalentHeroZone` flag-gating).
- **Lint:** clean (`--max-warnings=0`). **Build:** OK. **tsc:** 160 = baseline (**delta 0**); changed files clean.

## Notes
- `CreateTalentDialog` was included beyond the original spec ("consider") because it's the primary *new-agency* drift surface — same flag, `createTalent` already trims.
- The PopoverContent carries the 5d Escape-guard (`onEscapeKeyDown` stopPropagation), matching `LocationPicker`.

### DIFF
--- a/scripts/normalize-agency-names.ts
+++ b/scripts/normalize-agency-names.ts
@@ -1,0 +1,196 @@
+/**
+ * One-time fixup: normalize talent agency names to canonical values.
+ *
+ * Pass 1 — DUMP (dry-run, no mapping): print every distinct agency + count so
+ * the canonical map can be curated by hand from real data:
+ *   npx tsx scripts/normalize-agency-names.ts --clientId=unbound-merino
+ *
+ * Pass 2 — PREVIEW (dry-run, with mapping): show the planned renames:
+ *   npx tsx scripts/normalize-agency-names.ts --clientId=unbound-merino --mapping=agency-map.json
+ *
+ * Pass 3 — WRITE: apply the reviewed mapping:
+ *   npx tsx scripts/normalize-agency-names.ts --clientId=unbound-merino --mapping=agency-map.json --write
+ *
+ * The mapping JSON is a flat object of { "<existing variant>": "<canonical>" }.
+ * Match is on the trimmed, whitespace-collapsed value; unmapped agencies are
+ * left untouched (and reported). Map a variant to "" to clear the field.
+ *
+ * Run this BEFORE the AgencyCombobox flag flips: the agency filter is
+ * exact-equality (talentFilters.ts matchesAgency), so variants only collapse
+ * in the filter once the stored data is normalized.
+ */
+
+import { readFileSync } from "node:fs"
+import { getApp, getApps, initializeApp, applicationDefault } from "firebase-admin/app"
+import { FieldValue, getFirestore } from "firebase-admin/firestore"
+
+const PROJECT_ID = process.env.GCLOUD_PROJECT ?? "um-shotbuilder"
+
+function ensureApp() {
+  if (getApps().length) return getApp()
+  return initializeApp({ credential: applicationDefault(), projectId: PROJECT_ID })
+}
+
+/** Trim + collapse internal whitespace — the key both the dump and the mapping match on. */
+function normalizeKey(value: string): string {
+  return value.trim().replace(/\s+/g, " ")
+}
+
+function parseArgs(): { clientId: string | null; mappingPath: string | null; write: boolean } {
+  const args = process.argv.slice(2)
+  let clientId: string | null = null
+  let mappingPath: string | null = null
+  let write = false
+  for (const arg of args) {
+    if (arg.startsWith("--clientId=")) clientId = arg.slice("--clientId=".length).trim()
+    else if (arg.startsWith("--mapping=")) mappingPath = arg.slice("--mapping=".length).trim()
+    else if (arg === "--write") write = true
+  }
+  return { clientId, mappingPath, write }
+}
+
+/** Load + validate the reviewed { variant -> canonical } mapping, keyed on normalizeKey. */
+function loadMapping(path: string): Record<string, string> {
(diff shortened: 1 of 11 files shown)


ted-design:
Pushed `2cc49761` addressing the bot feedback:

**Codex P2 — collapse agency whitespace before persisting → FIXED.** `AgencyCombobox` now collapses internal whitespace (not just trims) in `commit`, in the `hasExactMatch` check, and on the "Add new" row. `IMG  Models` now matches the known `IMG Models` (no "Add new" row) and any new value is persisted whitespace-collapsed, so the exact-equality `matchesAgency` filter can't end up with near-duplicates. Added two tests covering this.

**claude-review 🟡 (both non-blocking) — dispositioned:**
- *`extractUniqueAgencies` case-sensitivity*: intentionally left as-is. Case canonicalization is the migration's job (Ted curates the `{variant → canonical}` map), and `normalizeKey` deliberately doesn't lowercase to preserve canonical casing ("IMG" not "img"). The concrete whitespace gap is now closed; case variants on a curated roster are unlikely and the filter/`hasExactMatch` already compare case-insensitively.
- *Inline CSS-var refs on the trigger*: these mirror the pre-existing flag-off `InlineEdit` pattern (`text-[var(--color-text-muted)]`) — CSS-var tokens, not hex; not a violation.

Also folded in flag-coverage tests for `featureTalentAgencyCombobox` (default-off + env parse, matching every other flag) and a `CreateTalentDialog` agency gating test.

Gates re-run locally on `2cc49761`: full suite green (3757 pass / 79 pre-existing skips), lint clean, build OK, tsc 160 = baseline (delta 0).

@codex review

ted-design:
Pushed `8281702c` for the claude-review follow-ups:

- **Item 1 (filter/exact-match normalisation mismatch) → FIXED.** `filtered` now collapses internal whitespace on the known agency too, matching `hasExactMatch`, so a pre-migration double-spaced agency stays selectable for a collapsed query. Added a test.
- **Item 5 (Clear visible while searching) → FIXED.** The "Clear agency" row is now hidden when a search query is active (`value && !typed`), so it can't be hit by accident below the results. Added a test.

Non-blocking items, dispositioned:
- **Item 2** (`createAgency || null` round-trip) — confirmed clean, no change.
- **Item 3** (`next === raw` unchanged guard) — confirmed correct, no change.
- **Item 4** (`aria-expanded`) — Radix `PopoverTrigger asChild` forwards `aria-expanded`/`aria-haspopup` to the button (visible in the test DOM dump: `aria-expanded="true" aria-haspopup="dialog"`). No change; will confirm in the flag-on `:5174` eyeball.
- **Item 6** (`__add__:` value prefix) — won't fix: a collision needs a real agency literally named `__add__:…`, and with `shouldFilter={false}` cmdk doesn't use the value for filtering. Noted.

Gates re-run locally on `8281702c`: full suite green (3759 pass / 79 pre-existing skips), lint clean, build OK, tsc 160 = baseline.

PR #466, opened 2026-06-16
feat(talent): casting matcher = its own ranked surface (flag-gated)

ted-design:
## Talent redesign — Phase 3 (LOCKED Decision #1)

Promotes the casting brief matcher from an in-place roster re-sort **mode** to its own score-first **surface**, behind a new build-time flag `featureCastingMatcherSurface` (env `VITE_CASTING_MATCHER_SURFACE`, default **OFF**; mirrors `featureTalentAgencyCombobox`).

### What changes
- **Flag ON:** when the casting brief is open, it renders as the **sole surface** (brief form + ranked results) replacing the roster. `castingMode` is forced off → the roster never re-sorts in place and never shows score badges. Built as a panel that can later graduate to a route.
- **Flag OFF:** the verbatim trunk behavior (brief panel above the re-sorting roster). **Byte-identical.**

### Diff note (please review with “hide whitespace”)
The `LibraryTalentPage.tsx` diff looks large but the **true functional change is 21 insertions / 1 deletion** — confirm with:
```
git diff -w main...feat/casting-matcher-surface -- src-vnext/features/library/components/LibraryTalentPage.tsx
```
The rest is a +4 re-indent of the **unchanged** flag-off block, now nested under the surface conditional. The flag-off JSX shows as unchanged context under `-w`, which is the proof of byte-identical flag-off.

### Blast radius — contained
`CastingBriefPanel` / `ScoreBadge` / `rankTalentForBrief` are imported **only** by `LibraryTalentPage`. This PR does **not** touch `AdminTalentDetailSheet` (board), the public `TalentDetailSheet` (`/s/:token`), or `CastingBoardPage` — so the §8 shared-component blast radius does not fire and **no casting-board write** is added. `rankTalentForBrief` (engine) and `CastingBriefMatcher` are unchanged.

### Tests
- `flags.test.ts`: 3 cases for the new flag (default-off, `1`/`true` enables, other values stay off).
- **New** `LibraryTalentPage.castingSurface.test.tsx` (flag-ON): roster shown by default → opening casting **swaps** to the dedicated surface (roster hidden) → closing returns to the roster.
- `LibraryTalentPage.test.tsx` (flag-OFF): opening casting keeps the **panel above the roster** (roster stays visible) — the discriminating coexistence assertion.

### Gates (local)
- ESLint: clean · `tsc`: 160 (= baseline, delta 0) · full `vitest`: **282 files / 3766 passed, 79 pre-existing skips, 0 failures**.

### Adversarial verify (5 read-only lenses on the diff)
flag-off-byte-identical **clean** · hook-deps/state **clean** · regression/blast-radius **clean**. Deferred-with-tracking (non-blocking, both for Ted's flag-on `:5174` eyeball):
1. **Flag-on integration coverage of the ranked cards** (matcher producing `MatchCard`s from a real requirement) is left to the eyeball — driving the Radix slider in jsdom is flake-prone, and the ranked path is fully covered at the unit level (`castingMatch.test.ts`, `CastingBriefMatcher.test.tsx`) and wired identically flag-on/off.
2. **Detail Sheet can stay layered over the surface** if a talent was selected before opening casting (functional; nav + close behave). An IA judgment call — whether entering the surface should also dismiss the Sheet — to decide at the eyeball.

### DIFF
--- a/src-vnext/features/library/components/LibraryTalentPage.tsx
+++ b/src-vnext/features/library/components/LibraryTalentPage.tsx
@@ -161,9 +161,7 @@ export default function LibraryTalentPage() {
     return m
   }, [castingResults])
 
-  // Flag-off keeps the legacy in-place re-sort mode. Flag-on (Phase 3) routes
-  // ranking to the dedicated matcher surface, so the roster never re-sorts or
-  // shows score badges.
+  // Flag-on: the matcher is a dedicated surface, not an in-place roster re-sort.
   const castingMode = !matcherSurface && castingPanelOpen && castingHasRequirements
   const castingSurfaceActive = matcherSurface && castingPanelOpen
 
--- a/src-vnext/features/library/components/LibraryTalentPage.tsx
+++ b/src-vnext/features/library/components/LibraryTalentPage.tsx
@@ -83,6 +83,7 @@ export default function LibraryTalentPage() {
   const canCreate = canManageTalent(role)
   const canEdit = canCreate && !isMobile
   const rosterIA = isFeatureEnabled("featureTalentRosterIA")
+  const matcherSurface = isFeatureEnabled("featureCastingMatcherSurface")
   const { data: talent, loading, error } = useTalentLibrary()
   const { data: projects } = useProjects()
   const projectIds = useMemo(() => projects.map((p) => p.id), [projects])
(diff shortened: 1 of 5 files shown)


ted-design:
@codex review

ted-design:
**Dispositions — pushed `7bc118d7`:**

**claude-review (required fix) — DONE.** Collapsed the 3-line `castingMode` comment to one task-free line per CLAUDE.md.

**Codex P2 "make ranked match cards selectable" — deferred to Ted's flag-on `:5174` eyeball (not built this PR), rationale:**
- The locked visual contract (`mockups/p9-casting-brief-matcher.html`) renders match cards as **non-interactive** informational rows, and `MatchCard` is non-interactive in flag-off today too. Adding click-to-detail is a new interaction model that isn't in the locked design — a design/eyeball decision reserved for Ted per the standing delegation.
- It would require threading selection into the **shared** `CastingBriefMatcher`/`MatchCard`, which renders in **both** flag states — risking the flag-off byte-identical guarantee that is the load-bearing property of this PR.
- The flag is OFF in prod (no live impact). The matcher is a scan/scope surface; acting on a candidate (open detail / casting board) lives on the roster/board by design.

Tracked as a flag-on eyeball item alongside the detail-Sheet-layering decision — both to be settled before the flag is promoted to default-on. If Ted wants click-through at the eyeball, it's a small, well-scoped follow-up.

PR #487, opened 2026-06-22
feat(export): persist shot-report config + reportType (R2 spine, flag-gated)

ted-design:
## R2 — persistence spine (PR-A of 2)

Part 1 of the export **reimagine R2** (config + recipe persistence). This PR is the **invisible persistence spine**; the visible surface (shot-report list, recipe picker, `looksMode` config UI) lands in the follow-up **PR-B** off `main` after this merges. All behind `featureShotReport` (OFF in trunk).

### What it does
Adds a `reportType` discriminator + a persisted `config` blob to the existing `useExportReports` Firestore layer (`clients/{c}/projects/{p}/exportReports`) so a saved **shot-report** doc coexists with legacy **block-canvas** docs in the same collection.

- **No migration.** A missing `reportType` reads as `"block-canvas"` (default applied in *both* `mapReport` and `loadReport`), so every pre-R2 doc is untouched.
- `createShotReport` (`setDoc`, `createdBy == uid`) + `saveReportConfig` (`updateDoc` — preserves `createdBy`/`pages`/`name`).
- `ShotReportPage` hydrates `config` from `?reportId=` and debounce-persists user edits. Hydration uses `setConfig` directly, so it never writes back what it just loaded.

### The load-bearing safety fix
The legacy block builder (`ExportBuilderPage`) auto-selects the most-recently-updated report on mount and a debounced auto-save writes its `pages` back. A shot-report doc (`pages: []`) reaching that path would render blank **and get clobbered to an empty block canvas**. Fix: `ExportBuilderPage` filters its report list to `reportType === 'block-canvas'` — applied at *every* consumer (auto-select init effect, delete fallback, `ExportTopBar` feed), so a shot-report id can never reach `setActiveReportId`. This is the **one always-on change**, and it's behavior-equivalent (all existing docs map to `block-canvas`).

### Deferred to PR-B (tracked, not parked)
`createShotReport` has **no UI caller in this PR** — its caller is PR-B's shot-report list / "new from recipe" flow. It ships here as the persistence API, exercised by the rules test (exact write shape) + unit tests. Owner: this initiative; phase: PR-B.

### Rules
**No `firestore.rules` change.** `exportReports` (rules:915–927) has zero field-level validation — `reportType`/`config` write straight through; the existing `createdBy` invariant is satisfied by the create/update paths. Adds the first-ever emulator rules test for this collection.

### Test plan
- `CI=1 npx vitest run` — `useExportReports.map.test.ts` (discriminator default-to-legacy + compat filter), `reportTypes.test.ts` (config JSON round-trip + default-merge), `reportModel.test.ts` (unchanged, green). **12/12.**
- Emulator: `exportReports.rules.test.ts` — producer create (createdBy==uid) ✓, config update preserves createdBy ✓, createdBy-spoof / foreign-tenant / viewer / anon all denied. **6/6.**
- `npm run typecheck:baseline` → 160/160 (0 new). `npm run build` ✓. eslint clean.

### DIFF
--- a/src-vnext/features/export/__tests__/exportReports.rules.test.ts
+++ b/src-vnext/features/export/__tests__/exportReports.rules.test.ts
@@ -128,6 +128,11 @@ describeOrSkip("firestore.rules — exportReports (shot-report docs)", () => {
       }),
     )
   })
+
+  it("[8] UPDATE that reassigns createdBy is denied (owner is immutable)", async () => {
+    const db = authed("prod-b", CLIENT_A, "producer")
+    await assertFails(updateDoc(reportRef(db, REPORT_OWNED), { createdBy: "prod-b" }))
+  })
 })
 
 // Visible skip notice so developers know why zero rules tests ran locally.
--- a/src-vnext/features/export/__tests__/exportReports.rules.test.ts
+++ b/src-vnext/features/export/__tests__/exportReports.rules.test.ts
@@ -118,6 +118,16 @@ describeOrSkip("firestore.rules — exportReports (shot-report docs)", () => {
   it("[6] anon CREATE fails", async () => {
     await assertFails(setDoc(reportRef(anon(), "report-anon"), shotReportDoc("nobody")))
   })
+
+  it("[7] same-tenant producer can UPDATE another producer's report (reports are tenant-editable; createdBy preserved)", async () => {
+    const db = authed("prod-b", CLIENT_A, "producer")
+    await assertSucceeds(
+      updateDoc(reportRef(db, REPORT_OWNED), {
+        config: { groupBy: "gender", excludedShotIds: [] },
+        updatedBy: "prod-b",
+      }),
+    )
+  })
 })
 
 // Visible skip notice so developers know why zero rules tests ran locally.
--- a/src-vnext/features/export/__tests__/exportReports.rules.test.ts
+++ b/src-vnext/features/export/__tests__/exportReports.rules.test.ts
@@ -1,0 +1,130 @@
+// @vitest-environment node
+/**
+ * Firestore rules tests for exportReports — the collection that now also holds
+ * saved shot-report docs (reportType:'shot-report' + config). This collection
+ * had ZERO rules coverage; these lock in the guarantees R2's writers rely on.
+ *
+ * Tested on a PRIVATE project so the permissive project wildcard catch-all is
+ * off (producerCanAccessProject requires visibility=='team'), leaving the
+ * explicit exportReports block — including its createdBy invariant — governing.
+ *
+ * Requires the Firestore emulator; self-skips when FIRESTORE_EMULATOR_HOST is unset.
+ */
+
+import { readFileSync } from "node:fs"
+import { resolve } from "node:path"
+import { afterAll, beforeAll, describe, expect, it } from "vitest"
+import {
+  assertFails,
+  assertSucceeds,
+  initializeTestEnvironment,
+  type RulesTestEnvironment,
+} from "@firebase/rules-unit-testing"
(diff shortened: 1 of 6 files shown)


ted-design:
Thanks both — addressed in `157edac3`:

**Fixed**
- **Stale config (Codex P2 + claude #5):** hydration effect now resets `config` to default when `?reportId=` is cleared, and cancels any pending debounced write when `reportId` changes — so an in-app nav `?reportId=X` → bare route (or A→B within 800ms) can't leak the previous report's grouping/exclusions or fire a stale write.
- **claude #1:** dropped the no-op `?? undefined` on `loadReport`'s `config`.
- **claude #2:** `.catch` on the debounced `saveReportConfig` call (matches `ExportBuilderPage`'s existing auto-save pattern — no unhandled rejection; richer save-status UX lands with the report UI in the surface PR).
- **claude #4:** renamed `ReportReportType` → `ExportReportType`.
- **claude #6:** added rules test `[7]` — a same-tenant producer can update another producer's report with `createdBy` preserved, pinning the intended tenant-editable behavior. (7/7 emulator rules tests green.)

**Skipped (rationale)**
- **claude #3 (magic-string default settings):** `createReport`/`importReport`/`loadReport` all already inline the same `{portrait, letter, Inter}`; there's no shared `DEFAULT_PAGE_SETTINGS` constant being bypassed, so extracting one is a separate cross-file refactor, out of scope for this spine PR. `createShotReport` matches the established pattern.

Context: this is the persistence **spine** (PR-A); `createShotReport`'s caller (the shot-report list + "new from recipe") lands in the follow-up surface PR. Gates re-verified locally: 12 unit + 7 emulator rules, tsc 160/160, build, eslint.

@codex review

ted-design:
Round 2 addressed in `2a0b1960`:

- **Codex P2 (hydration race):** `ShotReportPage` now tracks the hydrated `reportId` and won't persist a config edit until the active report has finished loading — so a quick toggle during the load window can't save the outgoing report's config onto the one being loaded.
- **claude #1 (test gap):** rules test `[8]` — an UPDATE reassigning `createdBy` is denied (owner immutable). Emulator rules now **8/8**.

claude #2 (bare `config` cast) and #3 (silent guard) were noted by the reviewer as fine-as-is / consistent with existing `settings` casting and the deferred save-status UI — leaving as-is for this spine PR; a config validation layer is a sensible future hardening when more knobs land.

Considering this converged — substantive items across both rounds are resolved; remaining notes are polish/future-hardening. Gates green locally: 12 unit + 8 emulator rules, tsc 160/160, build, eslint. Will merge on green CI.

@codex review

ted-design:
Round 3 addressed in `dce9dbfc`:

- **claude #1 (settings drift):** shot-report docs render from `config`, not block-canvas `PageSettings` — dropped the `settings` field from `createShotReport` entirely (`loadReport` has a fallback if ever needed). Removes the duplicated default rather than re-pointing it at a constant.
- **claude #2 (empty createdBy):** `createShotReport` now throws when unauthenticated instead of writing `createdBy:""`, which the create rule can never satisfy.

**Intentionally not changing:**
- **#4 (unvalidated reportType):** defaulting only the *missing* case is deliberate. Coercing an *unknown* string to `block-canvas` (the suggestion) would expose a future third report type to the very auto-select/clobber risk this PR prevents. An unknown type staying out of *both* lists is the safe failure mode; a real third type should be added to the union + its own filter.
- **#3 / #5:** the `saveReportConfig` guard mirrors the existing `saveReport` and is already `.catch`'d; user-facing save/load error UX lands with the report surface in PR-B.

Both bots confirmed the load-bearing pieces (the `blockCanvasReports` clobber-safety filter and the hydration write-back guard) are correct. Considering this converged — merging on green CI.

PR #491, opened 2026-06-23
feat(export): Product Info report (R4 PR1, flag-gated)

ted-design:
## R4 PR1 — Product Info report (the product-centric crew deliverable)

First of the two new R4 report types. A **product-centric** report — one card/section per product family in the project — applying the shipped shot-report **pattern** (one pure model → DOM + @react-pdf adapters → parity), **not** a new `ReportLayout` variant of the shot model. Behind **`featureProductInfoReport`** (default OFF, env `VITE_PRODUCT_INFO_REPORT`); **flag-off is byte-identical** to trunk.

### What's in it
- **`productInfoModel.ts`** — `deriveProductInfoModel(data, config)` (the inverse of `deriveShotReportModel`: walk non-deleted shots → looks → products → in-use families). Per family: colours/sizes used, `sizePending`, `isHero`, and `appears` (sorted shot numbers + look labels + status). Scope `in-use | library`, group-by `gender | product-type | none`, S/M/L image size. Reuses (not duplicates) `reportModel` helpers (now named exports) + the model-agnostic `resolveReportImages`.
- **`ProductInfoReportView.tsx`** (DOM, **one red = the hero mark**) + **`reportPdfProductInfo.tsx`** (landscape @react-pdf, lazy-imported on export; DOM ↔ PDF share the model). Explicit per-card pagination so a card never straddles/clips a page break.
- **List + report pages**, flag-gated routes (`export/product-reports`, `export/product-report`) + breadcrumbs + nav injection.
- **Persistence**: widen `ExportReportType` (+`product-info`) via a shared `createTypedReport`; reuse `saveReportConfig`. **No `firestore.rules` change** (rules validate nothing on `reportType`/`config`).

### Clobber-safety (scout-verified, decisive)
The new `reportType` is **safe by construction**: the only list-all consumer (`useExportReports` snapshot) never auto-acts, and every auto-acting surface filters by exact-match include (`ExportBuilderPage` → `block-canvas`, `ShotReportListPage` → `shot-report`), so a `product-info` doc is auto-excluded and can't be auto-selected/clobbered. The new list page filters `=== "product-info"` with explicit user actions only.

### Verification
- Built via a staged writer→adversarial-verify workflow (5 lenses: flag-off byte-identical, model-correctness, one-red+parity, clobber-persistence, wiring). **0 actionable findings**; 2 minors fixed (DOM print-preview clip-robustness, removed an unused `--sb-red-ink` token), rest accepted with rationale.
- Gates: **tsc baseline 160/160 · vitest 100/100** (incl. 14 new `productInfoModel` tests, +3 flags, breadcrumbs/nav coverage) **· lint · vite build** (`@react-pdf` stays in its own lazy chunk).

### Notes
- Two shipped shot-report files got a 1-line `as ReportConfig` narrowing each (forced by widening `config` to a union) — pure type narrowing, runtime-identical.
- `sizes` deliberately omits pending-scoped sizes (the comp surfaces "Size pending" as a distinct state); the `gender` group reuses the shared `GROUP_ORDER` (its `Mixed` bucket is filtered out — products only normalize to W/M/?).

**Flag stays OFF.** Eyeball gate is on `:5174` (with PR2/Talent) before any prod flag flip.

https://claude.ai/code/session_011YRgXcNvEuGVF7jwftCmB4

### DIFF
--- a/src-vnext/app/routes/breadcrumbs.ts
+++ b/src-vnext/app/routes/breadcrumbs.ts
@@ -113,6 +113,20 @@ export const breadcrumbsConfig: Record<string, BreadcrumbResolver> = {
     },
     { label: "Shot Report" },
   ],
+  "/projects/:id/export/product-reports": (ctx) => [
+    ...projectCrumbs(ctx),
+    { label: "Product Info" },
+  ],
+  "/projects/:id/export/product-report": (ctx) => [
+    ...projectCrumbs(ctx),
+    {
+      label: "Product Info",
+      to: ctx.params["id"]
+        ? `/projects/${ctx.params["id"]}/export/product-reports`
+        : undefined,
+    },
+    { label: "Product Report" },
+  ],
   "/projects/:id/schedules/:scheduleId/onset": () => [],
 
   "/requests": () => [{ label: "Shot Requests" }],
--- a/src-vnext/app/routes/index.tsx
+++ b/src-vnext/app/routes/index.tsx
@@ -61,6 +61,12 @@ const ShotReportPage = lazy(
 const ShotReportListPage = lazy(
   () => import("@/features/export/components/report/ShotReportListPage"),
 )
+const ProductInfoReportPage = lazy(
+  () => import("@/features/export/components/report/ProductInfoReportPage"),
+)
+const ProductInfoReportListPage = lazy(
+  () => import("@/features/export/components/report/ProductInfoReportListPage"),
+)
 const OnSetViewerPage = lazy(
   () => import("@/features/schedules/components/OnSetViewerPage"),
 )
@@ -290,6 +296,34 @@ export function AppRoutes() {
                   <ProjectScopeProvider>
                     <RequireDesktop label="Shot report">
                       <ShotReportPage />
+                    </RequireDesktop>
+                  </ProjectScopeProvider>
+                </RouteBoundary>
+              }
+            />
+          )}
+          {isFeatureEnabled("featureProductInfoReport") && (
+            <Route
+              path="projects/:id/export/product-reports"
+              element={
+                <RouteBoundary featureName="Product info reports">
+                  <ProjectScopeProvider>
+                    <RequireDesktop label="Product info reports">
+                      <ProductInfoReportListPage />
+                    </RequireDesktop>
+                  </ProjectScopeProvider>
+                </RouteBoundary>
+              }
+            />
+          )}
+          {isFeatureEnabled("featureProductInfoReport") && (
+            <Route
+              path="projects/:id/export/product-report"
+              element={
+                <RouteBoundary featureName="Product info report">
+                  <ProjectScopeProvider>
+                    <RequireDesktop label="Product info report">
+                      <ProductInfoReportPage />
                     </RequireDesktop>
                   </ProjectScopeProvider>
                 </RouteBoundary>
--- a/src-vnext/features/export/components/report/ProductInfoReportListPage.tsx
+++ b/src-vnext/features/export/components/report/ProductInfoReportListPage.tsx
@@ -1,13 +1,10 @@
 import { useCallback, useMemo, useState } from "react"
 import { useNavigate, useParams } from "react-router-dom"
 import { toast } from "sonner"
-import { Copy, FileText, Plus, Trash2 } from "lucide-react"
+import { Package, Plus, Trash2 } from "lucide-react"
 import { useAuth } from "@/app/providers/AuthProvider"
-import { isFeatureEnabled } from "@/shared/lib/flags"
-import { badgeVariants } from "@/ui/badge"
 import { Button, buttonVariants } from "@/ui/button"
 import { Input } from "@/ui/input"
-import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "@/ui/select"
 import {
   AlertDialog,
   AlertDialogAction,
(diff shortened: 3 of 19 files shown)


ted-design:
Thanks both — converged in `cd70aca3`. Disposition:

**claude-review**
1. **Hero flag wrong for SKU-level heroes** ✅ fixed — `matchesHeroProductId` was private in `lookHeroes.ts`; exported it and the model now uses the 3-way sku/colour/family match (was inherited from `resolveProducts`' family-only check). +test for a SKU-id `heroProductId`.
2. **"Appears in N shots" inflated** ✅ fixed — reshaped `ProductInfoAppearance` to **one entry per (shot, family)** carrying its look labels, so the count is distinct shots (also handles a family across Primary+Alt, and unnumbered shots). +test asserting 1 appearance for two same-family SKUs in one look.
3. **PDF fixed header stale range on overflow** ✅ fixed — dropped the per-sheet numeric range from the `fixed` header (kept the group label, which stays correct on any overflow page; footer page numbers remain dynamic).
4. **Export errors swallowed** ✅ added an export-failure toast on the Product Info page. (The identical gap in the shipped `ShotReportPage` is left as-is — out of scope here.)
5. **Double `findIndex` scan** ✅ single scan per anchor.

**Codex**
- **Count unique shot appearances once (P2)** ✅ same fix as claude #2.
- **Fall back to legacy `shot.products` (P2)** — **deferred, with rationale.** The in-use walk deliberately mirrors the shipped Shot Report's definition (`looks[].products` only) so the two reports describe the *same* product set; unioning the legacy flat `shot.products` here would make Product Info show products the Shot Report doesn't. It's an additive completeness enhancement, not a correctness bug, and both reports share the limitation today. Flagged for the `:5174` eyeball — if Q2-26 No. 3 (or a real project) has legacy shots with root-level products and no looks, I'll add the union as a follow-up across the report family.

Gates re-run green: tsc baseline 160/160, vitest 107/107, lint, vite build.

ted-design:
Round 2 converged in `c8fdab08`.

**Fixed (confirmed bugs + the perf/reuse items):**
1. **`imagesLoading` stuck when candidates → 0** ✅ reset it in the zero branch.
2. **Soft-deleted families in in-use scope** ✅ the in-use branch now filters `f.deleted !== true`, matching the library scope.
3. **Cross-type `reportId` clobber** ✅ added a `reportType` guard to **both** `ProductInfoReportPage` and `ShotReportPage` (never hydrate/arm write-back from a doc whose type doesn't match the page). My PR newly makes product-info docs clobberable via the shot page, so guarding both is the complete fix — and it also closes the pre-existing block-canvas↔shot-report exposure. Pure safety guard; a legit same-type doc always passes (full export suite green).
4. **`imageSize` spurious image-resolution cycle** ✅ the image effect now skips re-resolve when the candidate set is unchanged (an S/M/L toggle no longer flashes the overlay).
5. **Duplicate `resolveSrc`** ✅ reuse the shared `reportShared.resolveSrc`.

**Dispositioned (mirror the shipped report pattern — skip with rationale):**
- **ControlBar should use a shadcn primitive** — it mirrors `ReportView`/`ProductionSheetReport`'s existing hand-rolled control bars exactly. Extracting a shared `SegGroup`/`ToggleGroup` primitive is a worthwhile cross-cutting refactor, but it should land once across *all* report control bars, not be introduced asymmetrically in one new page. Deferred.
- **`--sb-red: #EB1400` "hardcoded hex"** — that line is the brand-red **token definition** (consumers use `var(--sb-red)`), exactly as the shipped `reportStyles.ts` + `reportPdfShared.ts` define it; the print `#fff` is intentional pure-white paper, matching the shipped print styles.
- **`pendingDelete` reset** — mirrors `ShotReportListPage`; the AlertDialog's `onOpenChange(false)` resets it on the normal close path. Left consistent with the sibling.

**Codex** re-raised the legacy `shot.products` fallback (P2) — disposition unchanged: deferred for report-family parity + flagged for the `:5174` eyeball.

Gates re-run green: tsc baseline 160/160, vitest 366 passed/8 pre-existing-skipped, lint, vite build.

ted-design:
Round 3 converged in $(git rev-parse --short HEAD).

- **Export fires while images still loading → no-image PDF** ✅ `imagesLoading` is threaded into the view; the Export button is now `disabled={exporting || imagesLoading}`.
- **Cross-type guard discarded edits on a deleted/missing doc** ✅ added an explicit `if (!full) return` before the `reportType` check on both pages, so a not-found doc and a cross-type doc are distinct (both correctly keep defaults / don't arm write-back).
- **Redundant `order` array in product-type grouping** ✅ removed — sorting the Map keys is identical.

**Codex**: only the legacy `shot.products` P2 re-raised — disposition unchanged (deferred for report-family parity; flagged for the `:5174` eyeball). No new items.

Gates green: tsc baseline 160/160, vitest 366 passed/8 pre-existing-skipped, lint, vite build.

ted-design:
Converged — round 4 reports **no blocking issues**. The five non-blocking findings, dispositioned:

1. **`new Set(config.excludedFamilyIds)` null guard** — not an actual crash: `new Set(undefined)`/`new Set(null)` both return an empty Set, the config type requires the array, and hydration always default-merges `DEFAULT_PRODUCT_INFO_CONFIG` (`excludedFamilyIds: []`). It mirrors the shipped `reportModel.ts` (`new Set(config.excludedShotIds)`, also unguarded). Left consistent with the sibling.
2. **Double Map read in `walkInUse`** / 3. **Masthead second pass** — micro-perf over ≤~35 families; readability favours the current form. Accepted.
4. **Private OKLCH palette in `productInfoStyles.ts`** — same as the shipped `reportStyles.ts`: a self-contained injected stylesheet scoping its own `--sb-*` tokens under `.sb-pir-root`. Consistent with the report family.
5. **File comments explain WHAT not WHY** — trimmed toward intent where it helped during the review; kept terse per CLAUDE.md.

Merging dark (flag OFF) per Ted's go. The `:5174` eyeball (with PR2/Talent, flags on) is the gate before any prod flag flip.

PR #492, opened 2026-06-23
feat(export): Talent report (R4 PR2, flag-gated)

ted-design:
## R4 PR2 — Talent report (the call-sheet-adjacent crew deliverable)

The second R4 report type. A **talent-centric** report — one card per `TalentRecord` in the project — applying the same pattern as the just-merged Product Info report (one pure model → DOM + @react-pdf adapters → parity), as its own `reportType`. Behind **`featureTalentReport`** (default OFF, env `VITE_TALENT_REPORT`); **flag-off is byte-identical** to trunk.

### What's in it
- **`talentModel.ts`** — `deriveTalentModel(data, config)`: **in-shots** scope (union of `shot.talentIds` over non-deleted shots, resolved against `data.talent`) or **project-attached** (`talent.projectIds` includes the project); **soft-deleted talent dropped in both scopes**. Per talent: `buildDisplayName`, `genderDisplayLabel`, agency, contact (email/phone/web), labeled measurements (`formatLabeledMeasurements`), headshot candidate, and `appears` = **one entry per shot** (looks accrued; count = distinct shots). `onHold` (any on-hold appearance) is the one decisive red.
- **`TalentReportView.tsx`** (DOM, **one red = the HOLD flag**) + **`reportPdfTalent.tsx`** (landscape @react-pdf, lazy; DOM ↔ PDF share the model). Initials-tile fallback for no headshot.
- **List + report pages**, flag-gated routes (`export/talent-reports`, `export/talent-report`) + breadcrumbs + nav. Config `groupBy: none|gender|agency`, `talentScope: in-shots|project-attached`.
- **Persistence**: widen `ExportReportType` (+`talent`) via `createTalentReport` over the shared `createTypedReport`. **No `firestore.rules` change.** Clobber-safe (the list filters `reportType === "talent"`).

### Inherits every PR1 fix
Built from the merged Product Info files as the template, so it carries the hardened patterns by construction (a dedicated verify lens confirmed this): cross-type `reportId` guard (`!full` + `reportType`), Export disabled while images load, image-key skip-reresolve, shared `resolveSrc`, export-error toast, no stale PDF header, per-shot appearance shape, soft-delete filter.

### Verification
- Staged writer→adversarial-verify workflow (5 lenses incl. a PR1-regression check): **all PASS, 0 actionable findings** (2 cosmetic cleanups already applied — a dead pluralization ternary, a misleading nav var name).
- Gates: **tsc baseline 160/160 · vitest 391/8-skip** (16 new `talentModel` tests, +3 flags, breadcrumbs/nav) **· lint · vite build** (`@react-pdf` lazy chunk).

**Flag stays OFF.** The `:5174` eyeball (both reports, flags on, throwaway project) is the gate before any prod flag flip.

https://claude.ai/code/session_011YRgXcNvEuGVF7jwftCmB4

### DIFF
--- a/src-vnext/app/routes/breadcrumbs.ts
+++ b/src-vnext/app/routes/breadcrumbs.ts
@@ -127,6 +127,20 @@ export const breadcrumbsConfig: Record<string, BreadcrumbResolver> = {
     },
     { label: "Product Report" },
   ],
+  "/projects/:id/export/talent-reports": (ctx) => [
+    ...projectCrumbs(ctx),
+    { label: "Talent" },
+  ],
+  "/projects/:id/export/talent-report": (ctx) => [
+    ...projectCrumbs(ctx),
+    {
+      label: "Talent",
+      to: ctx.params["id"]
+        ? `/projects/${ctx.params["id"]}/export/talent-reports`
+        : undefined,
+    },
+    { label: "Talent Report" },
+  ],
   "/projects/:id/schedules/:scheduleId/onset": () => [],
 
   "/requests": () => [{ label: "Shot Requests" }],
--- a/src-vnext/app/routes/index.tsx
+++ b/src-vnext/app/routes/index.tsx
@@ -67,6 +67,12 @@ const ProductInfoReportPage = lazy(
 const ProductInfoReportListPage = lazy(
   () => import("@/features/export/components/report/ProductInfoReportListPage"),
 )
+const TalentReportPage = lazy(
+  () => import("@/features/export/components/report/TalentReportPage"),
+)
+const TalentReportListPage = lazy(
+  () => import("@/features/export/components/report/TalentReportListPage"),
+)
 const OnSetViewerPage = lazy(
   () => import("@/features/schedules/components/OnSetViewerPage"),
 )
@@ -324,6 +330,34 @@ export function AppRoutes() {
                   <ProjectScopeProvider>
                     <RequireDesktop label="Product info report">
                       <ProductInfoReportPage />
+                    </RequireDesktop>
+                  </ProjectScopeProvider>
+                </RouteBoundary>
+              }
+            />
+          )}
+          {isFeatureEnabled("featureTalentReport") && (
+            <Route
+              path="projects/:id/export/talent-reports"
+              element={
+                <RouteBoundary featureName="Talent reports">
+                  <ProjectScopeProvider>
+                    <RequireDesktop label="Talent reports">
+                      <TalentReportListPage />
+                    </RequireDesktop>
+                  </ProjectScopeProvider>
+                </RouteBoundary>
+              }
+            />
+          )}
+          {isFeatureEnabled("featureTalentReport") && (
+            <Route
+              path="projects/:id/export/talent-report"
+              element={
+                <RouteBoundary featureName="Talent report">
+                  <ProjectScopeProvider>
+                    <RequireDesktop label="Talent report">
+                      <TalentReportPage />
                     </RequireDesktop>
                   </ProjectScopeProvider>
                 </RouteBoundary>
--- a/src-vnext/features/export/components/report/TalentReportListPage.tsx
+++ b/src-vnext/features/export/components/report/TalentReportListPage.tsx
@@ -67,11 +67,14 @@ export default function TalentReportListPage() {
 
   const handleDelete = useCallback(
     async (reportId: string) => {
+      setBusy(true)
       try {
         await deleteReport(reportId)
         toast.success("Report deleted")
       } catch {
         toast.error("Couldn't delete the report")
+      } finally {
+        setBusy(false)
       }
     },
     [deleteReport],
--- a/src-vnext/features/export/components/report/TalentReportListPage.tsx
+++ b/src-vnext/features/export/components/report/TalentReportListPage.tsx
@@ -1,0 +1,171 @@
+import { useCallback, useMemo, useState } from "react"
+import { useNavigate, useParams } from "react-router-dom"
+import { toast } from "sonner"
+import { Plus, Trash2, Users } from "lucide-react"
+import { useAuth } from "@/app/providers/AuthProvider"
+import { Button, buttonVariants } from "@/ui/button"
+import { Input } from "@/ui/input"
+import {
+  AlertDialog,
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@/ui/alert-dialog"
+import { PageHeader } from "@/shared/components/PageHeader"
+import { EmptyState } from "@/shared/components/EmptyState"
+import { useExportReports } from "../../hooks/useExportReports"
+import { DEFAULT_TALENT_CONFIG } from "../../lib/report/talentTypes"
+
+// Saved talent reports for a project: create, open, and delete. Sits beside the
+// single report page; never touches the legacy block-canvas builder or the
+// shot-report / product-info lists. Flag-gated route. Mirrors
+// ProductInfoReportListPage.
+
+export default function TalentReportListPage() {
+  const { id: projectId } = useParams<{ id: string }>()
+  const { clientId } = useAuth()
+  const navigate = useNavigate()
+  const { reports, loading, createTalentReport, deleteReport } = useExportReports(
+    clientId,
+    projectId,
+  )
+
+  const talentReports = useMemo(
+    () => reports.filter((r) => r.reportType === "talent"),
+    [reports],
+  )
+
+  const [newName, setNewName] = useState("")
+  const [busy, setBusy] = useState(false)
+  const [pendingDelete, setPendingDelete] = useState<{ id: string; name: string } | null>(null)
+
+  const openReport = useCallback(
+    (reportId: string) =>
+      navigate(`/projects/${projectId}/export/talent-report?reportId=${reportId}`),
+    [navigate, projectId],
+  )
+
+  const handleCreate = useCallback(async () => {
+    setBusy(true)
+    try {
+      const id = await createTalentReport(
+        newName.trim() || "Untitled report",
+        DEFAULT_TALENT_CONFIG,
+      )
+      setNewName("")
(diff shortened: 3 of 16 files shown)


ted-design:
Converged in this push.

**Fixed:**
1. **Zero-page PDF when all talent excluded** ✅ Export is now disabled when no non-excluded talent remain (`canExport = some group has a non-excluded item`), so `@react-pdf` can't throw on an empty Document. *(The merged Product Info report has the same edge — I'll add the identical guard there in a one-line follow-up before any flag flip.)*
2. **Divergent `initials()` copies** ✅ both the DOM view and PDF now import `initials` from `@/features/library/components/talentUtils` (first+last word), so a talent's initials match across the roster, the report card, and the PDF.
3. **Unguarded concurrent delete** ✅ the delete handler now busy-locks (shares the existing `busy` state with create, disabling the trash + create buttons during a delete).

**Dispositioned:**
- **`project.client` shows `clientId`** — `Project` carries only `clientId` (no name field), and the shipped `reportModel.ts` + the Product Info model use the same `data.project?.clientId`. Resolving a display name needs a clients lookup not wired into `ExportData` — a cross-report enhancement, not PR2's scope. Flagged for the `:5174` eyeball (if the masthead client reads as a slug, it's a report-family follow-up).
- **`#fff` in print media** — intentional pure-white print paper, matching the shipped `reportStyles.ts` print rules.
- **Codex: legacy `shot.talent` fallback** — the model uses `talentIds` (the reliable id link) exactly as the report family does; the legacy `shot.talent` is a *names* array that can't reliably resolve to `TalentRecord`s without fuzzy matching. Deferred + flagged for the eyeball (parallels the Product Info legacy-`shot.products` disposition).

Gates green: tsc baseline 160/160, vitest (export+nav+flags) pass, lint, vite build.

ted-design:
Round 2:

**Fixed:**
1. **`imagesLoading` stranded on a same-key mid-resolve re-run** ✅ real bug (inherited from the Product Info key-compare optimization). The effect now gates apply/clear on `lastImageKeyRef.current === key` (latest-request token) instead of a per-run `cancelled` closure, so an in-flight resolve always clears the overlay even when a groupBy/exclude toggle re-runs the effect with the same image set. *(The merged Product Info report shares this — bundled into the pre-flag-flip follow-up with its zero-page guard.)*

**Dispositioned:**
- **2 / Codex — legacy `shot.talent` scope** — the model uses `talentIds` (the reliable id link) as the whole report family does; legacy `shot.talent` is a *names* array that can't resolve to records without fuzzy matching. Deferred + flagged for the eyeball.
- **3 — `onHold` from shot status, not casting status** — intentional per the approved comp + R4-DESIGN: the HOLD flag means "this talent is in an on-hold *shot* → confirm before locking the call." A casting/booking-status flag is a different signal (needs a castingBoard read) — eyeball decision if wanted.
- **4 — double-create on rapid Enter** — mirrors the merged Product Info / Shot Report list pages; `!busy` + immediate `setBusy` covers the common case, and a rapid double-Enter just makes two empty reports (no data loss).
- **5 — `paginate` duplicated across reports** — the per-report paginators differ in per-sheet density + group handling; a shared generic paginator is a cross-cutting refactor across shot/product/talent, not a PR2-only change.
- **6 — HANDOFF.md / CHECKPOINT.md not updated** — those files don't exist in this repo and weren't touched by R1–R3; project state lives in the vault and is updated at `/wrap`.
- **7 — `--sb-red: #EB1400`** — the brand-red token *definition* (consumers use `var(--sb-red)`), matching the shipped `reportStyles.ts` / `productInfoStyles.ts`.

Gates green: tsc baseline 160/160, vitest 322 passed/8 pre-existing-skipped, lint, vite build.

ted-design:
Round 3:

**Fixed:**
- **2 — unused `notes`** ✅ removed from the model/type/test (the approved comp doesn't render it; trivially re-added at the eyeball if wanted).
- **3 — blank print preview when all-excluded** ✅ the paged view now shows "All talent are excluded — nothing to print." (pairs with the Export-disabled guard from round 1).
- **5 — redundant `has(a.title)` in the PDF** ✅ dropped (the model guarantees a title).

**Dispositioned:**
- **1 — DOM print preview density (6/sheet) ≠ PDF (3/page)** — the in-app print preview and the `@react-pdf` export are different renderers; both are clip-safe (whole cards never straddle a break), but the exact cards-per-page and 2-vs-3-across layout is a visual-density call best made seeing the real renders. The merged Product Info report already differs the same way. I'll align both to one agreed density at the `:5174` eyeball — flagging it there rather than guessing a number now.
- **4 — `project.client` = `clientId`** — unchanged disposition (no client-name field in `ExportData`; cross-report; eyeball).
- **6 — `withTalentReportsNav` splice duplicated a third time** — extracting a shared `insertAfterAnchor` helper would touch the merged `withShotReportsNav`/`withProductInfoReportsNav` too; a cross-cutting nav refactor, not PR2.
- **7 — arbitrary CSS-var Tailwind in the list page** — mirrors the merged `ProductInfoReportListPage` (`--color-*` token classes); consistent with the sibling.

Gates green: tsc baseline 160/160, vitest 322 passed/8 pre-existing-skipped, lint, vite build.

ted-design:
Round 4:

**Fixed:**
1. **Image effect — unmount setState vs same-key stranding (the round-2↔round-4 tension)** ✅ The two suggestions contradict: round 2 flagged the per-run `cancelled` flag *stranding* `imagesLoading` on a same-key re-run (so I removed it for a latest-key token); round 4 flags that with no `cancelled`, the resolve fires setState after unmount. Re-adding `cancelled` would reintroduce the round-2 bug. Resolved with an **unmount-scoped `mountedRef`** (set false only on unmount, not per-run) **plus** the key token — `if (mountedRef.current && lastImageKeyRef.current === key)`. A same-key re-run still lets the in-flight request clear the overlay; an unmount blocks the dead-component setState. Handles both.
2. **`loadReport` silent catch dropping edits** ✅ now toasts "Couldn't load the report — changes won't be saved" so a failed load doesn't silently discard edits.
4. **`STATUS[a.status]` unchecked index** ✅ `?? STATUS.todo` fallback, matching the DOM's `statusMeta` fallback (defensive against an out-of-enum Firestore status).

**Dispositioned:**
- **3 — `URL.revokeObjectURL` synchronous after `a.click()`** — pre-existing in **all four** PDF generators (`generateProductInfoPdf`/`generateShotReportPdf`/`generateExportPdf`/this one); PR2 copies the shipped pattern. A deferred-revoke fix should land across all generators at once, not asymmetrically in talent. Noting for a shared follow-up (alongside the Product Info zero-page + stuck-loading items already queued before the flag flip).

Gates green: tsc baseline 160/160, vitest 322 passed/8 pre-existing-skipped, lint, vite build.

ted-design:
Round 5:

**Fixed:**
1. **Nav label collision** ✅ the project "Talent" nav item is renamed **"Talent Reports"** (it shared a label + `users` icon with the Library `/library/talent` entry). Nav tests updated.
3. **Zero-page PDF on a direct `generateTalentPdf` call** ✅ `TalentPdfDocument` now renders a single "No talent to report." page when there are no non-excluded entries, so `@react-pdf` never gets a zero-page Document (the UI already disables Export via `canExport`).
4. **Shared `busy` cross-contaminating create/delete** ✅ reverted the delete busy-lock (added in round 1) so create and delete don't lock each other — **now matching the merged `ProductInfoReportListPage`**, which round 5 itself cites as correct. (Round 1 asked to add a delete guard; round 5 asked to remove/split it — reverting to the merged sibling resolves the two.)

**Dispositioned:**
- **2 — fluid view shows no message when all excluded** — respectfully, this is a misread: `GroupSection` maps **all** `group.items` (excluded ones render struck/dimmed with a Restore toggle); the "0 talent" is the printable count. Making `isEmpty` true on all-excluded would **hide the struck cards and break un-exclude on screen** — so I'm keeping the struck-card behavior (matches the merged Product Info report).
- **5 / 6 — blob-download + `paginate` duplicated across reports** — pre-existing across shot/product/talent; a shared download helper + generic paginator are cross-cutting refactors for the whole family (queued with the other pre-flag-flip follow-ups), not asymmetric PR2 edits.
- **7 — arbitrary Tailwind var in the list page** — mirrors the merged `ProductInfoReportListPage`.
- **8 — `withTalentReportsNav` encodes prior nav positions** — the anchor-priority (after Product Info → Shot Reports → Export) is inherent to ordering after siblings; a generic insert helper is the same cross-cutting nav refactor.
- **Codex — legacy `shot.talent`** — unchanged (talentIds is the reliable link; eyeball).

Gates green: tsc baseline 160/160, vitest 357 passed/8 pre-existing-skipped, lint, vite build.

ted-design:
Converged. Round 6 surfaces **no new correctness bugs** — every finding is a re-raise of an already-addressed item or a cross-cutting refactor of the merged report family. Disposition + merging dark (flag OFF) per Ted's go:

- **1 — `loadReport` failure blocks saves** — already mitigated round 4 with a toast (more than the merged shot/product pages, which are silent). A persistent banner / `saveBlocked` state is a larger UX treatment that should land consistently across the family, not asymmetrically here; the load-failure path is a rare transient.
- **2 — `isEmpty` on all-excluded** — unchanged: `GroupSection` renders excluded cards struck (reversible); hiding them would break un-exclude. Matches the merged Product Info report.
- **3 / 4 / 5 / 6 — blob-download · untagged config union · nav-injection helper · list-page structure** — all cross-cutting refactors across shot/product/talent (+ the hook). Real cleanups, but they touch merged code and belong in a report-family refactor, not an asymmetric PR2 edit. Logged for a follow-up.
- **7 — `ShotReportListPage.handleDuplicate` cast** — safe by construction: the duplicate source comes from the `reportType === "shot-report"`-filtered list, so `full.config` is always a `ReportConfig`; PR2 doesn't change that filter. Not worsened.
- **8 — raw report-chrome `<button>`** — mirrors the merged Product Info report's scoped report styling (the report intentionally uses its own brand chrome, not the app shadcn Button).
- **Codex — legacy `shot.talent`** — unchanged (talentIds is the reliable link; eyeball).

The cross-cutting items (shared list/nav/download helpers + a discriminated config union) are tracked for an optional report-family refactor. Gates green throughout: tsc 160/160, vitest, lint, build.

PR #494, opened 2026-06-24
feat(export): PR-B pre-flip bundle — report polish + Product Info hardening (flag-off)

ted-design:
Pre-flip **dark bundle** for the Product Info + Talent reports. All behind the existing flags (`featureProductInfoReport` / `featureTalentReport` / `featureShotReportRecipes`) — **prod byte-identical** until Ted flips them. Implements his ratified R4 `:5174` eyeball decisions plus the Product Info hardening parity that PR-A applied to the Talent report.

## Changes
- **D3 — remove the Talent HOLD flag entirely.** Confusing, and it rendered as a square red period (brand violation). Deletes `onHold` in lockstep across `talentTypes` → `talentModel` → `TalentReportView` (badge + masthead "Holds" stat) → `talentStyles` → `reportPdfTalent` → test, and cleans the now-orphaned `--sb-red` token. The **per-shot amber `on_hold` status dot** and the **Production Sheet's own HOLD flag** are different concepts and are untouched. Talent surface is now red-free.
- **#4 — Product Info image is colourway-first + deterministic.** New `pickColourwayFirstImage` (`skuImageUrl` → `familyImageUrl` → family thumbnail/header), hero-first. Stops preferring the ambiguous denormalized `thumbUrl`. A hero with no resolvable image no longer locks out a sibling that has one. The shot-report `pickProductImage` path is unchanged. *(Note: the product page's colourway data is sometimes wrong — that's a separate future data-quality job; this PR only makes the picker deterministic.)*
- **Brand — round dot.** Product Info hero mark square → round (DOM `border-radius:50%`, PDF `borderRadius:3` on a 6×6). The LIVE shot-report hero **bar** is left as-is.
- **D1 — title-case the client slug.** One shared `titleCaseSlug` at all 3 model boundaries (`unbound-merino` → `Unbound Merino`). Reaches the live shot report's client line too (approved).
- **D2 — align PDFs to the on-screen grid.** Product Info 4×3 (12/page), Talent 2×3 (6/page) wrapping grids. Image caps tightened (`IMAGE_MAX_HEIGHT` 232→96, `HEADSHOT_MAX_HEIGHT` 168→120). **⚠️ EYEBALL-GATED** — see below.
- **Hardening — Product Info.** Zero-page Export guard + `canExport` gate, and ports its image effect to the **StrictMode-safe `mountedRef` + latest-key** pattern (the bug that blanked Talent headshots in PR-A; Product Info had the same one).
- **Talent PDF URL wrap.** `breakLongToken` inserts U+200B so long portfolio URLs wrap (PR-A's `overflow-wrap` fix was DOM-only; @react-pdf ignores CSS overflow).

## Verification
- **Gates:** tsc baseline 160/160 (0 new) · 68 report unit tests · vite build · eslint — all green.
- **Built via** a ground-truth audit workflow (7 agents) → implement → **adversarial-verify workflow (5 skeptics)**. The skeptics found one colourway hero-lock edge (a hero with no image could regress a sibling's image) — **fixed + pinned with a regression test**. D3 lockstep, the StrictMode port (byte-identical to the canonical Talent page), round-dot, D1, and U+200B were all independently confirmed.

## ⚠️ Eyeball gate (the one thing static analysis can't confirm)
The D2 PDF density needs a real export on a **dense** project (most products/talent; a family with many colourways in many shots; a lead in 12+ shots):
- **Good:** a full 4×3 / 2×3 grid stacks on one landscape sheet, no card clipping a page break.
- **Watch for:** rows 2–3 pushed onto sparse continuation pages (a dense card taller than ~⅓ page) → lower the image cap (try 80) or accept sparse pages; and for a 12+ shot lead, confirm the in-shots list doesn't clip the card bottom (pre-existing `wrap={false}` behavior, *improved* by the cap reduction here).

After the eyeball, the prod flag flips are Ted's call (`VITE_PRODUCT_INFO_REPORT` / `VITE_TALENT_REPORT` / `VITE_SHOT_REPORT_RECIPES` in `deploy-live.yml` both build blocks) → then R5 retires the block canvas.

> NB: the `preview` CI check fails on a broken firebase-login preview-channel auth (infra, not code) — ignore for merge; affects all PRs.

### DIFF
--- a/src-vnext/features/export/components/report/ProductInfoReportPage.tsx
+++ b/src-vnext/features/export/components/report/ProductInfoReportPage.tsx
@@ -40,6 +40,10 @@ export default function ProductInfoReportPage() {
   // The image candidate set last resolved — re-derivations that don't change it
   // (an S/M/L toggle) skip the resolve so the loading overlay doesn't flash.
   const lastImageKeyRef = useRef<string | null>(null)
+  // False after unmount — gates async image setState so a navigate-away mid-resolve
+  // can't update a dead component. Unmount-scoped (NOT per-run), so a same-key
+  // re-run mid-resolve still lets the in-flight request clear the loading overlay.
+  const mountedRef = useRef(true)
 
   // Hydrate config from a saved product-info doc when ?reportId= is present, else
   // reset to defaults. Switching reports (or clearing reportId) also cancels any
@@ -88,13 +92,20 @@ export default function ProductInfoReportPage() {
     [reportId, saveReportConfig],
   )
 
-  // Cancel a pending config write if we unmount inside the debounce window.
-  useEffect(
-    () => () => {
+  // Cancel a pending config write + block async setState if we unmount. Reset to
+  // true on (re)mount — a remount (incl. StrictMode's mount→unmount→mount) must not
+  // leave it false, or the in-flight image resolve's setState is silently dropped
+  // while the keyed re-run skips re-resolving → empty imageMap (no product images).
+  useEffect(() => {
+    mountedRef.current = true
+    return () => {
+      mountedRef.current = false
+      // Clear the resolved-image key too, so a remount re-resolves cleanly instead
+      // of relying on the prior in-flight resolve landing after remount.
+      lastImageKeyRef.current = null
       if (persistTimer.current) clearTimeout(persistTimer.current)
-    },
-    [],
-  )
+    }
+  }, [])
 
   const model = useMemo(() => deriveProductInfoModel(data, config), [data, config])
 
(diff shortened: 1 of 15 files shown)


ted-design:
Good catch (Codex P2) — fixed in `6fbf88f2`. The Product Info `noImage` placeholder was 150pt against the new 96pt image cap; I capped it to the `IMAGE_MAX_HEIGHT` constant so a missing-image card is never taller than an imaged one. Applied the same fix to the Talent renderer, where the `initials` tile (132pt) was likewise taller than the 120pt headshot cap. Both now track their cap constant, so the denser 4×3 / 2×3 grids stay consistent regardless of image completeness. The absolute caps remain eyeball-gated on a dense export.

ted-design:
Thanks — dispositions on the 6 items (fixes in `f389ce95`):

**Fixed:**
- **#3 + #5 (titleCaseSlug dup / multi-line docstring):** `titleCaseSlug` now delegates to the shared `humanizeLabel(slug.toLowerCase())` — no duplicated split/capitalize logic, one-line comment. Behavior is identical (all 8 unit tests still green; `humanizeLabel` capitalizes word-initials, the pre-`toLowerCase()` gives the tail-lowercasing).
- **#4 (breakLongToken altitude):** moved into `reportPdfShared.ts` (it's a generic @react-pdf wrap helper with no talent deps); `reportPdfTalent` imports it.
- **#6 (imageIsHero):** now a `Set<string>` (the `false` value was never written) — removes the `.set(id,false)` misuse path.

**Deliberately not changed:**
- **#1 (ShotReportPage StrictMode) — false positive.** Its image effect uses a per-effect-run `let cancelled` (a *fresh* closure each mount — React's canonical fetch guard) and has **no key-skip early-return**. StrictMode's two mounts get two separate `cancelled` bindings, so the second mount's resolve checks its own (false) flag and `setImageMap` runs — images are not suppressed. The PR-A bug was specifically a **persistent `useRef` mountedRef without reset-on-mount, combined with a key-skip early-return** (the early-return is what strands it). `ShotReportPage` has neither, so the simpler guard is correct; adding a persistent `mountedRef` here would *introduce* the exact fragility PR-A had to fix, not remove it.
- **#2 (PRINT_COLS/PRINT_ROWS dedup) — deferred.** Real maintainability hazard, but the suggested fix (export the consts from the PDF file) would pull `@react-pdf/renderer` into the **eager DOM chunk**, defeating the lazy-PDF split. The correct fix is a small @react-pdf-free shared layout module imported by both the view and the PDF for each report type — a clean separate refactor. The values match today; tracked as a follow-up.

Gates green after the fix: tsc baseline 160/160, 68 report unit tests, vite build, eslint.

PR #500, opened 2026-08-04
feat(export): report-config Phase A — rename + hide-by-status (flagged)

ted-design:
## What
Phase A of the report-configuration initiative surfaced by the S3 batch eyeball (2026-08-04). Spec + phasing: `Projects/Shot Builder/outputs/2026-08-04-report-config/report-config-spec.md`.

Two features, across **all three report types** (shot / product-info / talent), behind a **new default-off flag** `featureReportConfig` (`VITE_REPORT_CONFIG`):

- **R1 — rename after create.** `renameReport(id, name)` = a name-only top-level merge that preserves `config`/`pages`/`createdBy` (mirrors `saveReportConfig`; the `createdBy` invariant holds → **no rules change**). Pencil affordance + rename dialog on all 3 report list pages.
- **R3 — include/exclude by shot status.** Optional `hiddenStatuses` on each config (DEFAULT-merged; **byte-identical when empty**). Shot report filters shots directly; product-info/talent drop an item **only when *every* appearance is a hidden status** (never-shot library / project-attached items are kept). Shared "Hide statuses" control in all 3 report views.

## Safety
- **Additive only** — no `firestore.rules` change, no migration (`exportReports` has no field-level validation; missing config fields DEFAULT-merge at read).
- **Flag-off byte-identical** — the controls are gated behind `featureReportConfig` (default OFF); the model filter is inert while `hiddenStatuses` is empty (its only writer is the gated control). So merge changes nothing in prod until the flag is flipped.
- **Deliberate go-live** — to enable, set `VITE_REPORT_CONFIG=1` in `deploy-live.yml` (both build blocks). The controls then appear on whichever report surfaces are already live (shot report today) + on product-info/talent when their flags flip.

## Gates (independently re-run)
- `typecheck:baseline` — ✅ 160/160, no new errors
- vitest — ✅ 79 model tests pass; R3 filter mutation-verified (revert → specific test goes red)
- `eslint --max-warnings=0` — ✅ clean
- `npm run build` — ✅ green

Built via a scout → TDD → adversarial-verify dynamic workflow; the status-filter semantics + rename no-clobber were verified with constructed counter-cases.

### DIFF
--- a/src-vnext/features/export/__tests__/exportReports.rules.test.ts
+++ b/src-vnext/features/export/__tests__/exportReports.rules.test.ts
@@ -133,6 +133,16 @@ describeOrSkip("firestore.rules — exportReports (shot-report docs)", () => {
     const db = authed("prod-b", CLIENT_A, "producer")
     await assertFails(updateDoc(reportRef(db, REPORT_OWNED), { createdBy: "prod-b" }))
   })
+
+  it("[9] producer UPDATE of name (rename) succeeds, preserving createdBy (renameReport path)", async () => {
+    const db = authed("prod-a", CLIENT_A, "producer")
+    await assertSucceeds(
+      updateDoc(reportRef(db, REPORT_OWNED), {
+        name: "Renamed",
+        updatedBy: "prod-a",
+      }),
+    )
+  })
 })
 
 // Visible skip notice so developers know why zero rules tests ran locally.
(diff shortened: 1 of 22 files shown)


ted-design:
Addressed the P1 (persisted `hiddenStatuses` applying after a flag-off rollback) in 6776a9c0: the derive **input** is now gated at all three report pages, so with `featureReportConfig` off the model receives `{ ...config, hiddenStatuses: [] }` and flag-off is truly byte-identical — even if a report persisted a filter while the flag was on. The derives stay pure, so the model/rules tests are unchanged. Gates re-run green (typecheck 160/160, eslint, model tests, build).

PR #510, opened 2026-08-09
feat: wayfinding quick-wires — Schedules nav, honest Export target, Display sheet

ted-design:
## DO NOT MERGE until the Aug 8–12 deploy freeze lifts.

Three precisely-scoped wiring fixes — no redesign, matches existing patterns exactly.

### 1. Nav: Schedules entry
`/projects/:id/schedules` (`ScheduleListPage`) and the on-set viewer already existed as routes in `src-vnext/app/routes/index.tsx` but had no nav entry anywhere — reachable only by typing the URL. Added a project-scoped **Schedules** item to `nav-config.ts`, placed between Casting and Call Sheet (schedules are what you create/manage before building a Call Sheet from one). No `surfaceBadge` — the backing Firestore rule allows read to anyone with project access, matching Call Sheet/Export/Tags. The on-set viewer itself is intentionally not linked directly; it's reached from a schedule, per the brief.

### 2. Export button now points at the live report
`ShotListPage.tsx`'s Export button (header actions + bulk-action bar) routed to the superseded block-canvas `/export?preset=shot-list`. Both call sites now navigate to `/projects/:id/export/reports` (the shot-reports list route, confirmed against `routes/index.tsx`). The block-canvas nav entry itself is untouched.

### 3. Reconnected the dead Display control
`ShotListDisplaySheet.tsx` (edits card-view field visibility) was fully built but imported nowhere; `ShotListPage` destructured `setFields` from `useShotListState` and never used it. Added a **Display** trigger button to `ShotListToolbar.tsx`, visible **card view only** (the table column system is untouched), that opens the sheet wired to `fields`/`onFieldsChange`. `fields`/`setFields` now flow from `useShotListState` → `ShotListPage` → `ShotListToolbar`.

## Tests
- `nav-config.test.ts`: Schedules entry exists, points at `/projects/:id/schedules`, and sits before Call Sheet.
- `ShotListPage.test.tsx`: Export click navigates to `/projects/:id/export/reports` (not the old preset URL).
- `ShotListToolbar.test.tsx` (new): Display trigger renders in card view only, opens the sheet, and toggling a field calls `onFieldsChange` with that field flipped.

## Gates run
- `npm run typecheck:baseline` — clean (160/160 baselined, no new errors)
- `CI=1 npx vitest --run` on all touched/added test files — 75/75 passed
- `npx eslint <touched files> --max-warnings=0` — clean
- `npm run build` — clean


ted-design:
Fix-pass commit `4fb7da28` addresses the confirmed [high] finding: both Export call sites (header button + BulkActionBar) navigated unconditionally to `/projects/:id/export/reports`, but that route only exists behind `featureShotReport` (forced on only in `deploy-live.yml`). On preview/dev/`.env` the button dead-ended in a 404 via the wildcard route.

Fix: extracted a single `exportTarget` const, gated on `isFeatureEnabled("featureShotReport")`, falling back to the legacy `/export?preset=shot-list` route when off — mirrors the existing gating pattern in `AppShell.tsx`. Both call sites now read from it, so they can't drift independently again.

Tests: flipped the existing route-pin test to assert the flag-ON target explicitly, and added a companion test pinning the flag-OFF fallback URL (`/export?preset=shot-list`), so the regression is falsifiable in both directions.

Gates re-run clean: `typecheck:baseline`, `vitest` (77/77 across the 6 touched suites), `eslint --max-warnings=0`, `build`.

ted-design:
## Medium/low sweep from the adversarial review

1. **Display sheet stayed open across a view-mode flip.** The "1"/"2" keyboard shortcut is bound on `document` in `ShotListPage`, so it had no way to know the Display sheet was open — flipping to table view while it was open left the sheet rendering its orphaned "Table Columns" branch (no trigger, no card-view equivalent). `ShotListToolbar` now closes the sheet the moment `viewMode` leaves `"card"`. Mutation-tested: reverting the effect reproduces the original failure in `ShotListToolbar.test.tsx`.
2. **Export flag-off fallback** — checked first; already covered by `ShotListPage.test.tsx`'s "Export falls back to the legacy block-canvas export when featureShotReport is off" test. No gap, nothing added.
3. **Deleted the stale `bareSchedules` → `/callsheet` redirect in `NotFoundPage.tsx`.** `/projects/:id/schedules` was un-retired for real by 9801e4fa (`ScheduleListPage`, live in `routes/index.tsx`); the live route now shadows this branch entirely, so it was dead — and it actively contradicted the new Schedules nav entry, which points at `/schedules`, not `/callsheet`. Added `NotFoundPage.test.tsx` to pin the live-route behavior and the three redirects that are still legitimate (bare project, `/inbox`, `/library`).
4. **Added a `SheetDescription`** to `ShotListDisplaySheet`, matching the house pattern in `ScheduleEntryEditSheet.tsx:237`.
5. **Schedules now has its own icon** (`list-checks`) instead of reusing Call Sheet's `calendar-days` — the two sit adjacent in the collapsed icon rail and were visually identical. Added to the `NavItemIcon` union + `nav-icons.tsx` map; tested that the two entries resolve different icons (`nav-config.test.ts`) and render genuinely different SVG markup (`nav-icons.test.tsx`, new).
6. **Documented two DEFERRED product items** — deliberately not fixed here, both move to the export design track:
   - Bulk-select Export lands on the reports **list**, not a populated document — intent carry-through (e.g. `?create=shot-list`) is deferred.
   - `ShotDetailPageUnified.tsx:407` (now with a one-line comment at the call site) still exports to the block-canvas, not the shot-reports flow the list page's Export button resolves to when `featureShotReport` is on — split-brain acknowledged, stays until the export wayfinding pass reconciles the two targets.

### Gates run
- `npm run typecheck:baseline` — clean (160/160 baselined)
- `CI=1 npx vitest --run` on all touched/added test files — 107/107 passed
- `npx eslint <touched files> --max-warnings=0` — clean
- `npm run build` — clean

PR #511, opened 2026-08-09
fix: offer project assignment when an existing user's role changes

ted-design:
**DO NOT MERGE until the Aug 8–12 deploy freeze lifts.**

## Summary

`InviteUserDialog.tsx` does the right thing for new users: it sets role AND assigns projects (`ProjectAssignmentPicker` + `bulkAddProjectMembers` / the CF's `assignToProjects`). `UserRoleSelect.tsx` — the inline role dropdown on the Team roster, used for **existing** users — only called `updateUserRole`, which sets the global Auth claim and nothing else. Result: making an existing user crew/warehouse/viewer silently granted them access to zero projects, since those roles depend on explicit project membership rather than org-wide rules access.

This PR wires the same assignment step into the existing-user path:

- New `RoleProjectAssignmentDialog` opens automatically after a role change to **crew/warehouse/viewer**, reusing `ProjectAssignmentPicker` and the exact `bulkAddProjectMembers` write path `InviteUserDialog` already uses.
- Admin can **skip** — this never blocks the role change (which is already written by the time the dialog opens).
- An inline warning renders whenever the interaction would otherwise end with **zero** project memberships ("X can't see any projects yet…").
- Producer/admin role changes are unaffected — no dialog, since those roles get org-wide access via rules (matches existing `defaultRole` clamping behavior elsewhere).

**Scope discipline (per decision):** wiring fix only. Admins remain the only ones who can grant project access — no changes to `functions/` or `firestore.rules`.

## Files

- `src-vnext/features/admin/components/RoleProjectAssignmentDialog.tsx` (new)
- `src-vnext/features/admin/components/RoleProjectAssignmentDialog.test.tsx` (new)
- `src-vnext/features/admin/components/UserRoleSelect.tsx` (opens the dialog for project-scoped roles; existing update+toast flow unchanged)
- `src-vnext/features/admin/components/UserRoleSelect.test.tsx` (mocks `AuthProvider` + the new dialog; adds wiring coverage)

## Test plan

- [x] `npm run typecheck:baseline` — 160/160 baselined, no new errors
- [x] `CI=1 npx vitest --run` on touched files — 57/57 passing; full `src-vnext/features/admin` suite — 90/90 passing, no regressions
- [x] `npx eslint <touched files> --max-warnings=0` — clean
- [x] `npm run build` — succeeds
- [x] Manual coverage in tests: (a) dialog appears for crew/warehouse/viewer role change, not producer/admin; (b) memberships written via the same path invite uses (`bulkAddProjectMembers` with identical shape); (c) zero-membership warning renders and clears as assignments are toggled; skip closes without writing; write failure surfaces a toast.

### DIFF
--- a/.tsc-baseline.json
+++ b/.tsc-baseline.json
@@ -6,6 +6,12 @@
     "sample": "No overload matches this call."
   },
   {
+    "file": "src-vnext/features/admin/components/RoleProjectAssignmentDialog.tsx",
+    "code": "TS2769",
+    "count": 1,
+    "sample": "No overload matches this call."
+  },
+  {
     "file": "src-vnext/features/admin/components/UserDetailPanel.tsx",
     "code": "TS2769",
     "count": 1,
--- a/src-vnext/features/admin/components/InviteUserDialog.tsx
+++ b/src-vnext/features/admin/components/InviteUserDialog.tsx
@@ -2,7 +2,7 @@ import { useState, useEffect } from "react"
 import { z } from "zod"
 import { toast } from "sonner"
 import { useAuth } from "@/app/providers/AuthProvider"
-import { ROLE, roleLabel } from "@/shared/lib/rbac"
+import { ROLE, roleLabel, isProjectScopedRole } from "@/shared/lib/rbac"
 import { ROLE_DESCRIPTIONS } from "@/shared/lib/roleDescriptions"
 import { inviteOrUpdateUser, bulkAddProjectMembers } from "@/features/admin/lib/adminWrites"
 import { ProjectAssignmentPicker, type ProjectAssignment } from "./ProjectAssignmentPicker"
@@ -224,7 +224,7 @@ export function InviteUserDialog({ open, onOpenChange }: InviteUserDialogProps) 
         <ProjectAssignmentPicker
           assignments={projectAssignments}
           onChange={setProjectAssignments}
-          defaultRole={role === "admin" ? "producer" : role}
+          defaultRole={isProjectScopedRole(role) ? role : ROLE.PRODUCER}
         />
       </div>
     </ResponsiveDialog>
--- a/src-vnext/features/admin/components/RoleProjectAssignmentDialog.a11y.test.tsx
+++ b/src-vnext/features/admin/components/RoleProjectAssignmentDialog.a11y.test.tsx
@@ -1,0 +1,144 @@
+/// <reference types="@testing-library/jest-dom" />
+import { describe, it, expect, vi, beforeEach } from "vitest"
+import { render, screen, waitFor } from "@testing-library/react"
+
+/**
+ * Focused a11y regression guard for RoleProjectAssignmentDialog.
+ *
+ * AdminPage.a11y.test.tsx mocks UserRoleSelect away entirely
+ * (`vi.mock("./UserRoleSelect", ...)`), so this dialog — which only mounts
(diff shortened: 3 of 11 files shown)


ted-design:
Medium/low sweep from the adversarial review, verified against current source before implementing (findings are prior-session shorthand, not always literal):

1. **Reactivate flow zero-access gap (the big one)** — confirmed: `deactivateUser` (functions/index.js:809-822) deletes every project member doc for the user; `reactivateUser` only restores the Auth claim + status. Reactivating as crew/warehouse/viewer guaranteed zero project access. `UserDetailPanel.handleReactivate` now reuses `RoleProjectAssignmentDialog` for project-scoped roles, same as the `UserRoleSelect` role-change path. Tested in the new `UserDetailPanel.test.tsx` (7 cases: crew/warehouse/viewer open it, producer/admin don't, a failed reactivate doesn't, missing admin uid doesn't).

2. **Test theater on `defaultRole`** — confirmed: the mocked `ProjectAssignmentPicker` hardcoded `role: "crew"`. Mock now honors `defaultRole`; added a `newRole="warehouse"` case that asserts the member doc carries `role: "warehouse"`. Deleting `defaultRole={newRole}` in the dialog now reddens it.

3. **Dead-end Assign button** — folded `handleAssign`'s `clientId`/`addedBy` preconditions into `disabled`, and skip rendering the dialog when `user?.uid` is absent (both `UserRoleSelect` and the new `UserDetailPanel` usage). **One correction to the finding as written**: I used `assignments.length === 0` for the disabled check, not the existing `hasNoAssignments` var. `hasNoAssignments` counts existing + new memberships together (it's for the warning banner) — reusing it would re-enable the dead-end click for any user who already has project access but picked zero *new* projects this time, since `handleAssign` still bails on `assignments.length === 0` regardless of existing memberships. Added a test pinning this (`addedBy: ""` + a project picked → still disabled).

4. **A11y** — added `role="status"` to the amber warning, and a new focused `RoleProjectAssignmentDialog.a11y.test.tsx` (this repo hand-mirrors axe rules rather than running axe-core in unit tests — matches `AdminPage.a11y.test.tsx` / `ProductListPage.a11y.test.tsx`'s existing house pattern). Confirmed `AdminPage.a11y.test.tsx` does mock `UserRoleSelect` away, so this dialog was genuinely uncovered.

5. **Mount vs open state** — split `assignDialogRole`/`assignOpen` in `UserRoleSelect` and the new `reactivateAssignRole`/`reactivateAssignOpen` in `UserDetailPanel`, matching `ProjectAccessTab`'s `AddProjectMemberDialog` pattern, so the Radix exit animation plays. Re: the "resets assignments when reopened" test — kept it rather than dropping it: before this fix the parent unmounted/remounted a fresh instance every time, so that internal effect was dead code in production; after the split, a same-instance open→close→reopen (e.g. a second role change) is a real reachable path, so the test now guards something real.

6. **Shared `isProjectScopedRole`** — added to `shared/lib/rbac.ts`, with tests. `UserRoleSelect` derives its check from it. `InviteUserDialog`'s `role === "admin" ? "producer" : role` clamp and `UserDetailPanel`'s identical clamp are now both `isProjectScopedRole(role) ? role : ROLE.PRODUCER` — verified these are behaviorally identical for all 5 roles before swapping.

Gates: `typecheck:baseline` ✅ (161/161 baselined, no new errors), `eslint --max-warnings=0` on all touched files ✅, `npm run build` ✅, `vitest` on the full `src-vnext/features/admin` dir — 10 files / 107 tests, all green.

PR #514, opened 2026-08-09
feat(export): production sheet becomes the default shot-report recipe; style numbers render unbroken

ted-design:
## ⚠️ DO NOT MERGE until #513 is deployed and Ted gives the word

This PR is the code half of a two-part rollout. #513 (env-only) flips
`featureShotReportRecipes` ON in prod. Once that's live and confirmed, Ted
decides when this one merges — it changes the shot-report DEFAULT and needs
that flag live first, or the new default is invisible (flag-off always
clamps to image-led, by design — see below).

## Summary

**1. Default recipe (Ted's 2026-08-09 decision).** `DEFAULT_REPORT_CONFIG.layout`
is now `"production-sheet"` for brand-new shot reports (was `"image-led"`).
Two forward-compat floors keep this from leaking where it shouldn't:

- `resolveReportLayout(config, recipesEnabled)` — single source for all 3
  places layout gets resolved (ReportView's screen render, ShotReportPage's
  PDF export, ShotReportListPage's create/duplicate persist paths). When
  `featureShotReportRecipes` is OFF it **always** returns `"image-led"`,
  regardless of `config.layout` or the new default — so dev/preview (no env
  var) and any flag rollback keep the old behavior, and a report created
  while the flag is off can never persist the new default into a saved doc.
- `hydrateReportConfig(stored)` — floors an **absent** `layout` field to a
  frozen `LEGACY_REPORT_LAYOUT` ("image-led"), not the current default. A
  genuinely pre-R3 saved report (written before recipes existed) keeps
  rendering exactly what it always has; it does not silently "migrate" the
  day this ships. Existing saved reports that DO carry a persisted layout
  keep it, full stop — nothing migrates them.
- **Duplicate-report is a documented exception, by design**: `handleDuplicate`
  in `ShotReportListPage` copies the SOURCE report's `config` verbatim
  (layout included), even while `featureShotReportRecipes` is OFF — cloning a
  saved production-sheet report while the flag is off still writes
  `layout: "production-sheet"` into the new doc. This is intentional (a
  "duplicate" that silently downgraded the source's own recipe would be
  surprising), and it's safe: the *render* stays clamped to image-led by
  `resolveReportLayout` regardless of what's persisted, so nothing ships the
  new layout to a viewer while the flag is off — only the saved doc's config
  field carries it forward, ready to resume once the flag is on.

**2. Style numbers render unbroken.** Real-data renders showed every style
number with stray visible spaces — `"W- TP- BU- 1045"`, `"UM2025- 252"`.

Root cause (verified by rendering a real PDF via `renderToBuffer` and
inspecting it with PyMuPDF, not guessed): `breakLongToken` injected a literal
U+200B (zero-width space) between breakable characters (`/ . ? & = _ - @ + :
% # , ;`) so `@react-pdf` could wrap long SKUs/URLs. `@react-pdf`'s
non-embedded base-14 Helvetica/WinAnsi path can't encode U+200B — it silently
truncates the codepoint to its low byte (`0x200B & 0xFF = 0x0B`) and draws
the WinAnsi glyph living at `0x0B`, which carries the **same advance width as
a real space** (measured 1.807pt at 6.5pt Helvetica — exactly the space
glyph). Every breakable character printed a visible gap.

Fix: `hyphenateToken` / `tokenHyphenation` in `reportPdfShared.ts` now drive
`@react-pdf`'s `hyphenationCallback` prop instead of pre-processing the
string. `@react-pdf` rejoins the callback's returned parts byte-for-byte when
a line doesn't need to break, so **no glyph is ever inserted** for a typical
token (up to ~16 chars) — while a pathologically long token (the #508
overflow scenario) still gets real line breaks and doesn't blow its column.
Fixed at the shared root so production-sheet, balanced-rows, and the talent
report's contact values all pick it up.

## Review-pass fixes (this update)

An adversarial review of the above surfaced 3 HIGH, 2 MEDIUM, and 3 LOW
findings, all verified against the code and fixed:

- **H1** — `reportRecipeStyleTokenSpacing.test.tsx`'s "no can't-wrap warning"
  assertion was unfalsifiable: `warn.mockRestore()` ran *before*
  `warn.mock.calls` was read, and `mockRestore` clears recorded calls (same
  as `mockReset`), so the assertion passed unconditionally regardless of
  whether the warning fired. Fixed by moving the restore to `afterEach`,
  mirroring the existing `reportRecipeOverflow.test.tsx` pattern.
- **H2** — the flag-off clamp was only proven at the `resolveReportLayout`
  helper level, never at a real call site. Added: (a) a `ReportView` render
  test asserting `data-layout="image-led"` with the flag off even when
  `config.layout` carries `"production-sheet"`; (b) a `ShotReportListPage`
  create-flow test (real render + click) asserting the Firestore write's
  `config.layout` is `"image-led"`, not the new default.
- **H3** — `ShotReportPage`'s hydrate effect had an unclamped branch: a
  legacy doc with **no `config` field at all** (`full.config === undefined`)
  fell through to `setConfig(DEFAULT_REPORT_CONFIG)` directly, handing it the
  NEW `"production-sheet"` default; the next debounced save would persist
  that into what was previously an image-led-only doc. Fixed:
  `setConfig(hydrateReportConfig(loaded ?? {}))`, so the no-config case floors
  to `LEGACY_REPORT_LAYOUT` exactly like the missing-layout-field case
  already did. Added a real-render test: a no-config doc, flag off, click a
  config control, assert the persisted write's layout is `"image-led"`.
- **M1** — `hyphenateToken` kept the break character at the *end* of each
  syllable (`"W-"`, `"TP-"`, `"LS-"`, `"1066"`). `@react-pdf/textkit` draws an
  **unconditional** real `"-"` glyph at every syllable-boundary wrap
  (`breakLines` → `insertGlyph(..., HYPHEN, ...)`, no way to opt out) — so a
  wrap landing on a syllable that already ends in `"-"` doubles it
  (`"W--TP--LS--1066"` wherever a wrap actually falls), and the same
  mechanism inserts a spurious `"-"` right after an `"@"` or `"."` in a
  talent email/URL. Fixed by emitting the break character at the **head** of
  the *next* syllable instead (`"W"`, `"-TP"`, `"-LS"`, `"-1066"`), so a
  drawn hyphen and the token's own hyphen never land on the same physical
  line. Unit-tested directly (no syllable both ends AND the next starts with
  a breakable char) and falsifiable: reverting to tail-emission reddens it.
- **M2** — extended the real-render spacing suite with long-token coverage
  for `BalancedRowsPdfDocument` and `TalentPdfDocument` (a long agency email
  in a contact value). Also documented — empirically, not by assumption — why
  a blanket "no `--` anywhere in the flattened extracted text" check is a
  **false-positive trap** for any token that wraps across multiple physical
  lines: the extractor concatenates every line's text runs in stream order
  with no line-boundary awareness (by its own docstring, "NOT a general PDF
  parser"), so it reports a same-string `"--"` at *every* wrap regardless of
  whether the fix is applied or reverted (verified against both states: 5
  hits either way on the pathological fixture). The suite now uses that
  check only where it's actually valid — a token long enough to matter but
  still fitting on one line — and covers the wrapping case with what real
  rendering *can* prove (no overflow warning, no reintroduced U+000B, no
  content lost).
- **L1** — `useExportReports.ts`'s `"image-led"` fallback literal now imports
  `LEGACY_REPORT_LAYOUT` instead of duplicating the string.
- **L3** — removed the dead `?? "image-led"` fallback arm on
  `DEFAULT_REPORT_CONFIG.layout` in `ShotReportListPage` (the constant always
  carries a value at runtime; the interface just types the field optional
  for persisted blobs, which was making the fallback look load-bearing when
  it wasn't).

## Mutation-verified (not just green — proven falsifiable)

- Reverted `resolveReportLayout`'s flag branch → both flag-off clamp tests
  (including the new H2 call-site tests) went red; restored → green.
- Reverted `hydrateReportConfig`'s legacy floor → the pre-R3-blob test went
  red; restored → green.
- Reverted the two production-sheet/balanced-rows `Text` elements to the old
  `breakLongToken` string-injection + removed `hyphenationCallback` → all
  real-render regression tests in `reportRecipeStyleTokenSpacing.test.tsx`
  went red (confirmed via extracting the actual PDF content-stream text, not
  React props); restored → green.
- Reverted `hyphenateToken`'s head-emission back to tail-emission → the new
  unit test ("never draws two adjacent hyphens across a wrap point") went
  red; restored → green.
- Reverted the H3 fix to a bare `DEFAULT_REPORT_CONFIG` fallback → the new
  no-config-doc test went red; restored → green.

## Gate results

- `npm run typecheck:baseline` — ✅ no new type errors (160/160 baselined)
- `CI=1 npx vitest --run <touched + related test files>` — ✅ all green
  (full `src-vnext/features/export` suite also run for extra safety: 43
  files / 478 tests passed, 9 pre-existing skips, 0 failures)
- `npx eslint <touched files> --max-warnings=0` — ✅ clean
- `npm run build` — ✅ succeeds

## Test plan

- [ ] After #513 deploys and the flag is confirmed ON in prod, create a new
      shot report and confirm it defaults to the "On-set sheet"
      (production-sheet) recipe.
- [ ] Open an existing pre-recipes shot report and confirm it still renders
      image-led (unchanged).
- [ ] Export a production-sheet or balanced-rows PDF with a real style number
      (e.g. `W-TP-LS-1066`) and confirm it prints as one unbroken token.

### DIFF
--- a/src-vnext/features/export/components/report/ReportView.tsx
+++ b/src-vnext/features/export/components/report/ReportView.tsx
@@ -28,6 +28,7 @@ import {
   REPORT_LAYOUT_OPTIONS,
   REPORT_SORT_FIELD_OPTIONS,
   REPORT_STATUS_OPTIONS,
+  resolveReportLayout,
 } from "../../lib/report/reportTypes"
 import type { SortDir } from "../../lib/report/reportSort"
 import { hasAnyIncludedShot, sizeLabel } from "../../lib/report/reportModel"
@@ -649,10 +650,11 @@ export function ReportView(props: ReportViewProps): JSX.Element {
   const [printMode, setPrintMode] = useState(false)
 
   // Recipes ride their own flag; flag-off forces image-led so prod is byte-identical
-  // to the live R1/R2 report regardless of any persisted config.layout.
+  // to the live R1/R2 report regardless of any persisted config.layout (or of
+  // DEFAULT_REPORT_CONFIG.layout — resolveReportLayout ignores both when off).
   const recipesEnabled = isFeatureEnabled("featureShotReportRecipes")
   const reportConfigEnabled = isFeatureEnabled("featureReportConfig")
-  const layout: ReportLayout = recipesEnabled ? (config.layout ?? "image-led") : "image-led"
+  const layout: ReportLayout = resolveReportLayout(config, recipesEnabled)
 
   const toggleExclude = (shotId: string): void => {
     const set = new Set(config.excludedShotIds)
--- a/src-vnext/features/export/components/report/ShotReportListPage.tsx
+++ b/src-vnext/features/export/components/report/ShotReportListPage.tsx
@@ -59,7 +59,12 @@ export default function ShotReportListPage() {
   // Picker starts on the shipped default recipe (production-sheet as of the
   // 2026-08-09 decision) rather than a hardcoded "image-led" literal, so it
   // can't drift from DEFAULT_REPORT_CONFIG if the default changes again.
-  const [recipe, setRecipe] = useState<ReportLayout>(DEFAULT_REPORT_CONFIG.layout ?? "image-led")
+  // DEFAULT_REPORT_CONFIG.layout is always set at runtime — the interface
+  // types it optional (a persisted blob can lack it) but the shipped default
+  // constant always carries a value, so an "?? image-led" fallback here was
+  // dead code that quietly reintroduced the hardcoded literal this comment
+  // says to avoid.
+  const [recipe, setRecipe] = useState<ReportLayout>(DEFAULT_REPORT_CONFIG.layout!)
   const [busy, setBusy] = useState(false)
   const [pendingDelete, setPendingDelete] = useState<{ id: string; name: string } | null>(null)
   const [pendingRename, setPendingRename] = useState<{ id: string; name: string } | null>(null)
--- a/src-vnext/features/export/components/report/ShotReportListPage.tsx
+++ b/src-vnext/features/export/components/report/ShotReportListPage.tsx
@@ -32,6 +32,7 @@ import {
   DEFAULT_REPORT_CONFIG,
   REPORT_LAYOUT_LABEL,
   REPORT_LAYOUT_OPTIONS,
+  resolveReportLayout,
   type ReportConfig,
   type ReportLayout,
 } from "../../lib/report/reportTypes"
@@ -55,7 +56,10 @@ export default function ShotReportListPage() {
   const recipesEnabled = isFeatureEnabled("featureShotReportRecipes")
   const reportConfigEnabled = isFeatureEnabled("featureReportConfig")
   const [newName, setNewName] = useState("")
-  const [recipe, setRecipe] = useState<ReportLayout>("image-led")
+  // Picker starts on the shipped default recipe (production-sheet as of the
+  // 2026-08-09 decision) rather than a hardcoded "image-led" literal, so it
+  // can't drift from DEFAULT_REPORT_CONFIG if the default changes again.
+  const [recipe, setRecipe] = useState<ReportLayout>(DEFAULT_REPORT_CONFIG.layout ?? "image-led")
   const [busy, setBusy] = useState(false)
   const [pendingDelete, setPendingDelete] = useState<{ id: string; name: string } | null>(null)
   const [pendingRename, setPendingRename] = useState<{ id: string; name: string } | null>(null)
@@ -69,11 +73,14 @@ export default function ShotReportListPage() {
   const handleCreate = useCallback(async () => {
     setBusy(true)
     try {
-      // Recipes flag-off => always image-led, so the create config can't smuggle
-      // an unreviewed layout into prod.
+      // Recipes flag-off => always image-led (via the shared clamp, NOT a raw
+      // DEFAULT_REPORT_CONFIG spread — DEFAULT_REPORT_CONFIG.layout is now
+      // production-sheet, so spreading it verbatim would persist the new
+      // default into a doc created while the flag is off: dev/preview, or a
+      // flag rollback, would then smuggle an unreviewed layout into prod).
       const config = recipesEnabled
         ? { ...DEFAULT_REPORT_CONFIG, layout: recipe }
-        : DEFAULT_REPORT_CONFIG
+        : { ...DEFAULT_REPORT_CONFIG, layout: resolveReportLayout(DEFAULT_REPORT_CONFIG, recipesEnabled) }
       const id = await createShotReport(newName.trim() || "Untitled report", config)
       setNewName("")
       openReport(id)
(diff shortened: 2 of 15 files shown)


ted-design:
## Review-pass fixes applied (commit 5aef1870)

Applied all 8 adversarial-review findings against this PR — verified each
against the code before fixing, no findings skipped:

| # | Finding | Fix |
|---|---|---|
| H1 | `warn.mockRestore()` ran before `warn.mock.calls` was read (mockRestore clears recorded calls) — the "no can't-wrap warning" assertion was unfalsifiable | Moved restore to `afterEach`, mirroring `reportRecipeOverflow.test.tsx` |
| H2 | Flag-off clamp only proven at the `resolveReportLayout` helper, never at a real call site | Added `ReportView` render test + `ShotReportListPage` create-flow test (real render + click + write assertion) |
| H3 | `ShotReportPage`'s no-config hydrate branch called `setConfig(DEFAULT_REPORT_CONFIG)` directly, handing a legacy doc the NEW production-sheet default | `setConfig(hydrateReportConfig(loaded ?? {}))` — floors to image-led like the missing-`layout`-field case already did |
| M1 | `hyphenateToken` kept the break char at the END of each syllable; `@react-pdf/textkit` draws an unconditional real hyphen at every wrap, doubling it (`W--TP--...`) | Break char now emitted at the HEAD of the next syllable; unit-tested and falsifiable |
| M2 | Missing long-token coverage for `BalancedRowsPdfDocument` + `TalentPdfDocument` | Added, and documented (empirically) why a blanket "no `--`" check on the flattened extracted text is a false-positive trap for multi-line-wrapped tokens — the extractor has no line-boundary awareness. Used correctly-scoped checks instead. |
| L1 | `useExportReports.ts` had a duplicated `"image-led"` literal | Imports `LEGACY_REPORT_LAYOUT` |
| L3 | Dead `?? "image-led"` fallback arm in `ShotReportListPage` | Removed |
| L2 | Duplicate-report's by-design layout-preservation behavior undocumented | Documented in the PR body |

**Worth flagging on M2**: my first pass at the M2 tests asserted "no `--`
anywhere in the rendered text" on a pathologically long (multi-line-wrapping)
token for `BalancedRowsPdfDocument` — it failed. Investigated before treating
it as a real bug: rendered both the fixed and pre-fix code against the same
fixture and got 5 `"--"` hits either way, confirming it's an artifact of the
test's naive text extractor (concatenates separate physical lines with no
boundary awareness), not a rendering defect. Corrected the test to check
what's actually provable at the render level for a wrapping token (no
overflow warning, no reintroduced U+000B, all content present) and added a
same-line-fitting long-token case where the "--" check is valid.

Gates: `typecheck:baseline` clean, `eslint --max-warnings=0` clean on
touched files, full `src-vnext/features/export` suite green (43 files / 478
tests, 9 pre-existing skips), `npm run build` succeeds. Pushed to
`feat/export-production-sheet-default`.

PR #523, opened 2026-08-17
perf(shots): rAF-coalesce column resize; persist width on mouseup

ted-design:
## What / why

Plan item 0.2 (Shots Premium Surface build plan, Phase 0). Two related fixes to `useColumnResize`:

1. **rAF-coalesce `onWidthChange`.** `mousemove` fires far more often than the browser paints; `onWidthChange` is now coalesced to at most one call per animation frame, always flushing the last pending frame on `mouseup` so the true end-of-drag width is never lost.
2. **New optional `onResizeEnd(key, width)` callback**, fired once per drag on `mouseup`. `ShotsTable` uses it to defer the localStorage persist (`useTableColumns.setColumnWidth -> persist -> setItem`) to mouseup instead of writing on every frame — the live width is held in local state (`dragWidthOverride`) during the drag and only applied to the colgroup/header. Row cells don't render per-column pixel widths, so this override never touches `visibleColumns`, meaning it doesn't interact with row memoization (PR for 0.3) either.

Both changes are additive/backward-compatible: `onWidthChange` keeps its existing signature (same final value, just delivered less often), and `onResizeEnd` is optional — the other 4 consumers of `useColumnResize` pass nothing new and are unaffected.

## Blast radius

`useColumnResize` is shared by 5 tables: shots, `ProductFamiliesTable`, `LocationsTable`, `TalentTable`, `CallSheetCastTable`. Ran the sibling test files as a required gate:

- `src-vnext/features/schedules/components/CallSheetCastTable.test.tsx` — 8 tests, green
- `src-vnext/features/library/components/__tests__/TalentTable.test.tsx` — 10 tests, green
- `src-vnext/features/library/components/__tests__/LocationsTable.test.tsx` — 8 tests, green
- **`ProductFamiliesTable` has no test file** (confirmed via `find` — none exists in the repo). Not exercised by an automated gate; the change to `useColumnResize` is additive-only (new optional field in the options object, existing behavior preserved for callers that don't pass it), so no behavior change for this consumer, but flagging per the blast-radius protocol.

## Test + mutation evidence

New file: `src-vnext/shared/hooks/__tests__/useColumnResize.test.ts` (5 tests):
- 10 `mousemove`s fired within one (unflushed) animation frame → `onWidthChange` not yet called, exactly 1 frame scheduled; flushing the frame → exactly 1 call, with the correct FINAL width.
- clamping to `minWidth`/`maxWidth` still holds while coalescing.
- pending frame is flushed and `onResizeEnd` fires exactly once on `mouseup`, even when the rAF never got to paint.
- `onResizeEnd` reports the starting width when `mouseup` fires with no movement.
- omitting `onResizeEnd` doesn't throw (backward compatibility check).

**Mutation:** reverted `handleMouseMove` to call `onWidthChange` directly per mousemove (bypassing the rAF coalescing). Result: the coalescing test failed — 10 calls recorded instead of 1 (`expected "spy" to not be called at all, but actually been called 10 times`) — and the pending-flush-on-mouseup test also failed. Reverted the mutation; suite back to 5/5 green.

## Gates (this branch)

- `CI=1 npx vitest --run` on the new test file + blast-radius set — all green (see above)
- `npm run typecheck:baseline` — ✅ no new type errors (161/161 baselined)
- `npx eslint --max-warnings=0` on changed files — clean
- `npm run build` — succeeds

## Deviations

None. `useTableColumns.ts` was NOT touched — the persist-deferral is entirely local to `ShotsTable.tsx`'s consumption of `useColumnResize`.

### DIFF
--- a/src-vnext/features/shots/components/ShotsTable.tsx
+++ b/src-vnext/features/shots/components/ShotsTable.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useRef, type ReactNode } from "react"
+import { useCallback, useMemo, useRef, useState, type ReactNode } from "react"
 import {
   DndContext,
   closestCenter,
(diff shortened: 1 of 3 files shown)


ted-design:
**Fix: mouseup reported startWidth instead of dragged width (P1, adversarial review)**

**Defect:** on mouseup, `onResizeEnd?.(state.key, pending?.width ?? state.startWidth)` fired, but the drag's last rAF-coalesced `mousemove` had already flushed and nulled `pendingRef` before mouseup read it — true for essentially every real drag — so `onResizeEnd` (and thus the persisted column width) fell back to the pre-drag `startWidth`.

**Fix:** `handleMouseUp` now cancels any still-pending rAF (`cancelAnimationFrame`) before it can null the ref, captures `pendingRef.current` directly, flushes it explicitly via `onWidthChange`, then calls `onResizeEnd` with that captured width (falling back to `startWidth` only for a genuine no-movement drag). This is a different mechanism than a separate `lastWidthRef` but produces the same correct result with one fewer ref.

**Proven red on unfixed hook** (fix reverted, tests kept):
```
FAIL  flushes the pending frame and fires onResizeEnd exactly once on mouseup, even if the rAF never painted
  expected "spy" to not be called at all, but actually been called 1 times
  Received: 1st spy call: ["name", 240]
FAIL  reports onResizeEnd with the starting width when mouseup fires with no movement
  expected "spy" to be called 1 times, but got 0 times
Tests  3 failed | 2 passed (5)
```
Fix restored → all 5 tests green.

**Gates:** vitest (5/5 pass) · `typecheck:baseline` (161/161 baselined) · `eslint --max-warnings=0` (clean) · `npm run build` (succeeds).

Not merging — independently re-verified and pushed to `phase0-resize-throttle` for review.

PR #524, opened 2026-08-17
perf(shots): memoize table rows

ted-design:
## What / why

Plan item 0.3 (Shots Premium Surface build plan, Phase 0). `RowCells` and `SortableRow` in `ShotsTable.tsx` are wrapped in `React.memo` with a comparator scoped to `{shot, visibleColumns, isSelected, ctx}` — the only props that actually determine a row's rendered output. Everything else (`onOpenShot`, `stopPropagation`, `reorderDragHandle`, `lanes`, `onAssignScene`, `showLifecycleActions`, `renderLifecycleAction`) is table-wide config or a stable/curried callback, deliberately excluded from the comparator.

Without this, toggling ONE row's selection (or any other unrelated state change) re-renders every row's cells.

### Two supporting changes needed for the memoization to actually work

1. **`useTableColumns.visibleColumns` is a fresh array reference every call** — it's derived via `.filter().toSorted()`, uncached, so it's a brand-new array even when nothing about the columns changed. That silently broke the row comparator for every row on every unrelated re-render (confirmed empirically — first test pass showed all 33 cells across all 3 rows re-rendering, not just the toggled row). Row cells only care WHICH columns are visible and in what ORDER — never pixel width (that's colgroup/header-only) — so a local `stableVisibleColumns` is memoized off the column-key sequence and passed to rows; the live `visibleColumns` (with current widths) still drives the colgroup/thead directly.
2. **The per-row `onToggle` closure was built inline in ShotsTable's render loop** (`() => selection.onToggle(shot.id)`), a fresh closure every render for every row regardless of memoization. Replaced with `onToggleShot`, `selection.onToggle` forwarded as-is (unchanged reference); `RowCells` curries it with `shot.id` internally — which only happens when `RowCells` itself actually re-renders. `GroupRows` (grouped-view rendering, not memoized, out of scope for this fix) was updated to match since it shares `RowCellsProps`.

## Blast radius

**Local.** `RowCells`, `SortableRow`, `RowCellsProps`, `SortableRowProps` are all module-private to `ShotsTable.tsx`. No shared hook or component (`useTableColumns.ts`, `useColumnResize.ts`, `ResizableHeader.tsx`, etc.) is touched — `useTableColumns` is only *consumed* here, not modified. No blast-radius gate applies per the build plan's classification of this item.

Ran the broader shots suite anyway as a sanity check (not required, but cheap): `ShotListPage.test.tsx` (35), `ShotListPage.unifiedEditorFork.test.tsx` (2), `ShootShotList.test.tsx` (10) — 47/47 green, confirming the `onToggle` → `onToggleShot` prop rename didn't regress selection/toggle behavior anywhere else that renders `ShotsTable`.

## Test + mutation evidence

New file: `src-vnext/features/shots/components/__tests__/ShotsTable.rowMemo.test.tsx` (2 tests, one for the plain `<tr>` path and one for the `SortableRow`/DnD path). Spies on `renderShotCell` (called once per visible column per row) to record which shot id's cells actually re-rendered, then toggles one row's selection checkbox and asserts only that row's id appears in the post-toggle call list.

**Mutation:** unwrapped `React.memo(...)` from both `RowCells` and `SortableRow` (reverting to plain function components). Result: both tests failed — all 3 rows re-rendered (33 `renderShotCell` calls across ids `a`/`b`/`c` instead of just the toggled row's ~11). Reverted the mutation; suite back to 2/2 green.

## Gates (this branch)

- `CI=1 npx vitest --run` on the new test file — 2/2 green
- `npm run typecheck:baseline` — ✅ no new type errors (161/161 baselined)
- `npx eslint --max-warnings=0` on changed files — clean
- `npm run build` — succeeds

## Deviations

- `GroupRows` (grouped-view rendering) required a matching `onToggleShot` prop-wiring update since it shares `RowCellsProps` with the primary row path — it is NOT memoized itself (out of scope; grouped-view row unification is Phase 2 per the build plan) and its behavior is unchanged, just re-wired to the renamed prop.
- Added a `stableVisibleColumns` local memo (see above) to make the row comparator actually effective — this wasn't explicitly called out in the plan's 0.3 description but was necessary for the mutation-verified test to pass; without it, `visibleColumns`'s per-render fresh-array identity defeated memoization for every row, every render. Confined to `ShotsTable.tsx`; `useTableColumns.ts` itself is untouched.

### DIFF
--- a/src-vnext/features/shots/components/ShotsTable.tsx
+++ b/src-vnext/features/shots/components/ShotsTable.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useRef, useState, type ReactNode } from "react"
+import { memo, useCallback, useMemo, useRef, useState, type ReactNode } from "react"
 import {
   DndContext,
   closestCenter,
@@ -122,7 +122,12 @@ type RowCellsProps = {
   readonly visibleColumns: readonly TableColumnConfig[]
   readonly selectionEnabled: boolean
   readonly isSelected: boolean
-  readonly onToggle?: () => void
+  /**
+   * Stable across renders (it's `selection.onToggle`, forwarded as-is — never
+   * a per-row closure built by the caller). RowCells curries it with `shot.id`
+   * internally, which only happens when RowCells itself actually re-renders.
+   */
+  readonly onToggleShot?: (shotId: string) => void
   readonly showLifecycleActions: boolean
   readonly renderLifecycleAction?: (shot: Shot) => ReactNode
   readonly onOpenShot: (shotId: string) => void
@@ -133,12 +138,32 @@ type RowCellsProps = {
   readonly onAssignScene?: (shotId: string, laneId: string | null) => void
 }
 
-function RowCells({
+/**
+ * Only these four props determine whether a row's cells need to re-render.
+ * Everything else (onOpenShot, stopPropagation, reorderDragHandle, lanes,
+ * onAssignScene, showLifecycleActions, renderLifecycleAction, onToggleShot)
+ * is table-wide config or a stable/curried callback that doesn't vary
+ * independently of a data or selection change — see React.memo(RowCells) below.
+ */
+function rowCellsPropsAreEqual(prev: RowCellsProps, next: RowCellsProps): boolean {
+  return (
+    prev.shot === next.shot &&
+    prev.visibleColumns === next.visibleColumns &&
+    prev.isSelected === next.isSelected &&
+    prev.ctx === next.ctx
+  )
+}
+
+// Memoized: with N shots, a single unrelated re-render (e.g. toggling one
+// row's selection) would otherwise re-render every row's cells. The
+// comparator above lets React skip rows whose own data/selection/columns
+// haven't changed.
+const RowCells = memo(function RowCells({
   shot,
   visibleColumns,
   selectionEnabled,
   isSelected,
-  onToggle,
+  onToggleShot,
   showLifecycleActions,
   renderLifecycleAction,
   ctx,
@@ -160,7 +185,7 @@ function RowCells({
             checked={isSelected}
             onCheckedChange={(v) => {
               if (v === "indeterminate") return
-              onToggle?.()
+              onToggleShot?.(shot.id)
             }}
             aria-label={isSelected ? "Deselect shot" : "Select shot"}
           />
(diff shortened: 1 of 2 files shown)


ted-design:
**Fix: row-memo comparators omitted selectionEnabled/showLifecycleActions (P1, adversarial review)**

**Note on this branch's history:** the commit this PR previously carried (`perf(shots): memoize table rows`) added `React.memo` to `RowCells`/`SortableRow` but did **not** fix the reviewed defect — its comparator still only covered `{shot, visibleColumns, isSelected, ctx}`, with a comment asserting `showLifecycleActions`/`reorderDragHandle` were safe to omit. That assertion was wrong per the confirmed defect (they each gate a `<td>`), and the test it shipped with covered a different property (cross-row render isolation) than the one requested. This commit corrects the comparators in place; the prior commit's cross-row-isolation test and optimization are kept (they're valid, just not sufficient).

**Defect:** the comparators omitted `selectionEnabled` and `showLifecycleActions`, both of which gate an entire `<td>` in the row output, while `thead`/`colgroup` (unmemoized) update immediately. Toggling selection mode or lifecycle-actions visibility left memoized rows on their old `<td>` count, misaligning the table body by a column against the fresh header (live toggle at `ShotListPage:531`).

**Fix:** added `selectionEnabled`, `showLifecycleActions`, and (RowCells only — `SortableRowProps` omits `reorderDragHandle` from its type entirely, since `SortableRow` always builds its own via `useSortable`) `reorderDragHandle !== undefined` to both comparators, with a comment explaining why `lanes`/`renderLifecycleAction`/`onAssignScene`/`onToggleShot` are safely excluded — verified against the actual code: `lanes` is rescued via `rowContexts`' `laneById` dependency, carried on the already-compared `ctx`; the others are stable callbacks the row curries with `shot.id` internally.

**Proven red on unfixed comparators** (fix reverted, tests kept):
```
FAIL  keeps every row's <td> count in sync with <thead> after selectionEnabled toggles off
  row 0 <td> count: expected 13 to be 12
FAIL  keeps every row's <td> count in sync with <thead> after showLifecycleActions toggles on
  row 0 <td> count: expected 14 to be 13
Tests  2 failed | 2 passed (4)
```
Fix restored → all 4 tests green.

**Gates:** vitest (4/4 pass) · `typecheck:baseline` (161/161 baselined) · `eslint --max-warnings=0` (clean) · `npm run build` (succeeds).

Not merging — pushed to `phase0-row-memo` for review.

